PR #2, opened 2015-01-09
another attempt at s3 region redirecting

geemus:
I think this may fix https://github.com/fog/fog/issues/3275

See also: http://stackoverflow.com/questions/27091816/retrieve-buckets-objects-without-knowing-buckets-region-with-aws-s3-rest-api


Nitrodist:
The problem with this fix is that the error being raised in https://github.com/fog/fog/issues/3275 is `Excon::Errors::BadRequest` which is not listed in the rescue block above.

PR #29, opened 2015-01-20
Add PlacementTenancy to launch configuration parser and test case

benpillet:
Let me know if I should update any part to conform with coding standards etc.


benpillet:
I'd like to add a test case for this, but I'm not quite sure where.  Should it go in https://github.com/fog/fog-aws/blob/master/tests/requests/auto_scaling/model_tests.rb#L12  ??

geemus:
Good question. I'm not sure there is a particularly good place for it (and it should be a fairly safe change). We can discuss more, but for simplicity I'm going to just pull this in for now. Thanks!

PR #30, opened 2015-01-20
[DISCUSS] allow rspec as a testing framework.

lanej:
Per the discussion https://github.com/fog/fog/issues/1266, this is an example of what an rspec collection and model test would entail.

  The biggest issue I can see is that in order to facilitate shindo's 'one resource per suite', you have to do a ` before(:all)` (frowned upon) with an instance variable and then a subject reference to the instance variable.  In order to fit the more modern take on reseting state before each example, we could easily move the setup to a `before(:each)` but it would blow out the time for testing against a real provider.

  Huge wins are the tagging, the command line execution of examples via filters, and also the community familiarity.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,8 +1,16 @@
 require "bundler/gem_tasks"
+require 'rspec/core/rake_task'
 
 task :default => :test
 
 mock = ENV['FOG_MOCK'] || 'true'
-task :test do
-  sh("export FOG_MOCK=#{mock} && bundle exec shindont")
+
+task :test => ["test:shindo", "test:rspec"]
+
+namespace :test do
+  desc 'Run shindo tests'
+  task :shindo do
+    sh("export FOG_MOCK=#{mock} && bundle exec shindont")
+  end
+  RSpec::Core::RakeTask.new(:rspec)
 end
--- a/Rakefile
+++ b/Rakefile
@@ -1,9 +1,9 @@
 require "bundler/gem_tasks"
-require 'rspec/core/rake_task'
+require "rspec/core/rake_task"
 
 task :default => :test
 
-mock = ENV['FOG_MOCK'] || 'true'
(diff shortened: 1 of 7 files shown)


tokengeek:
There will be a number of MiniTest based specs as part of core that providers will need to include to make sure their implementations behave against these standard interfaces (shared examples as Minspec modules).

So in the future there will be a minimum of those that need running in the tests.

So just be aware those MUST be run, the Shindo test SHOULD be run and using RSpec mean there will be three different sets of tests for this.

Since `fog-aws` is not official but is the most popular we need to be careful we don't splinter the fog testing across numerous frameworks and in order to test you need many libraries to do the same thing (and slow down CI runs etc.)

Also since it is not official it's not as if there is an organisation with their own code testing rules that need to be followed.

We've not said modules can't use their own testing frameworks but I'm raising it as a concern of mine.

I would have preferred to have gone with RSpec to begin with but we went with Minitest/Spec so that's how the shared behaviour is being implemented.

plribeiro3000:
@tokengeek i share your concern. The reason we are moving from shindo is to make development easier to new devs (at least what i could understant), so if each provider has its own test framework it will not make sense at all.

lanej:
@tokengeek @plribeiro3000 ok.  I was unaware of the `fog-core` testing "interface" using minitest.  I understand that fracturing the testing frameworks could be really annoying as fog continues to be dissected.

I'll close this and re-submit with a minitest-based solution

plribeiro3000:
@lanej I think we can move forward with that discussion but it would be better to happen in the context of all providers instead of just one and once we  agree we could move together.

lanej:
@plribeiro3000 understood.

geemus:
I'm not opposed to rspec (though I tend to prefer minitest). We did already start down that path, but if it warrants revisiting perhaps we can/should. Seems I'm pretty out numbered.

PR #31, opened 2015-01-20
fix autoscaling activities collection setup

fcheung:
The AutoScaling#Group.activities method was doing something very strange - loading the collection rather than letting it load itself.

The activities model itself was also doing something very strange since it was assuming that all the arguments passed to it were filters, whereas we need to pass in the service (and since filters are dumped straight into the request parameters this caused an exception in the request signing code).

I'm not super familiar with the Fog::Collection stuff, but I think this is closer to what it should do (and does actually work).


geemus:
Yeah that does look weird before and better now. Thanks!

PR #36, opened 2015-01-22
simulate sns confirmation message

lanej:
* make mock account_id look more real
* expose region on Mock (already on Real)


geemus:
Thanks!

PR #37, opened 2015-01-22
default namespace and evaluation period on alarm

michelleN:
* deleted initialization method on cloud watch alarm
* filter alarms by alarm name in describe_alarms request

### DIFF
--- a/lib/fog/aws/models/cloud_watch/alarm.rb
+++ b/lib/fog/aws/models/cloud_watch/alarm.rb
@@ -27,6 +27,12 @@ module Fog
         attribute :threshold, :aliases => 'Threshold'
         attribute :unit, :aliases => 'Unit'
 
+        def initialize(attributes)
+          self.namespace ||= "AWS/EC2"
+          self.evaluation_periods ||= 1
+          super
+        end
+
         def save
           requires :id
           requires :comparison_operator
@@ -35,8 +41,6 @@ module Fog
           requires :statistic
           requires :threshold
 
-          self.namespace ||= "AWS/EC2"
-          self.evaluation_periods ||= 1
 
           options = Hash[self.class.aliases.map { |key, value| [key, send(value)] }]
           options.delete_if { |key, value| value.nil? }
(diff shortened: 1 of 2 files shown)


geemus:
@michelleN thanks! The filters for the mocks look good, but I'm a bit confused around the init/save changes. Could you describe why those values should move to the new location? Thanks!

michelleN:
@geemus When trying to pass in namespace as a symbol, the initializer was setting both :namespace and the alias. Two attributes were getting set (:namespace and ‘Namespace’) and only the attribute in the initializer was being taken into account.

The logic is that AWS sets default values for ‘evaluation_period’ and :namespace attributes, so we don’t really need an initializer method if we set up the default attribute values in the save method.

geemus:
Hmm. Could we accomplish the same end by leaving it in the init, but changing to use symbol keys there? (or using the setters like it is in save now). I wouldn't worry about it too much, but in other models I expect defaults to be there, so it might be easier to overlook inside of save. What do you think?

michelleN:
@geemus yes, I think that works too.

lib/fog/aws/models/cloud_watch/alarm.rb
```diff
@@ -28,21 +28,20 @@ class Alarm < Fog::Model
...
-          attributes['Namespace']         ||= 'AWS/EC2'
+          self.namespace ||= "AWS/EC2"
+          self.evaluation_periods ||= 1
           super
         end
 
         def save
           requires :id
           requires :comparison_operator
-          requires :evaluation_periods
```
geemus: Could we reinstate these also, just in case? Otherwise someone could set them to nil explicitly and run into weird errors. Thanks!

geemus:
Thanks, just one more small piece of feedback remaining.

lib/fog/aws/models/cloud_watch/alarm.rb
michelleN: absolutely. @geemus

geemus:
Thanks!

PR #39, opened 2015-01-24
Allow for AWS errors not specifying region

greysteil:
We're getting ```"undefined method `captures' for nil:NilClass"``` when sending requests with the wrong region. This PR updates the error parsing to handle the case where a new region isn't specified (the comment above this line suggests only some errors explicitly provide it).


geemus:
Are you seeing that behavior on master or an older version? I think the case that would cause it has been fixed, such that it might not even encounter this any more, but I'm not sure. Could you double check your version and if it still occurs? Thanks!

greysteil:
We're on 0.0.7

geemus:
Hmm. I guess it must be yet-another-edge-case then. Thanks for double checking.

PR #40, opened 2015-01-26
Add idempotent excon option to some route53 API calls

josacar:
This PR add `idempotent` attribute to underlying request calls to retry in case AWS throws a 400 error.

Affected API calls:

- ChangeResourceRecordSetsRequest
- GetHostedZone
- ListResourceRecordSets

### DIFF
--- a/lib/fog/aws/requests/dns/change_resource_record_sets.rb
+++ b/lib/fog/aws/requests/dns/change_resource_record_sets.rb
@@ -141,6 +141,7 @@ module Fog
           body = %Q{<?xml version="1.0" encoding="UTF-8"?><ChangeResourceRecordSetsRequest xmlns="https://route53.amazonaws.com/doc/#{@version}/">#{changes}</ChangeResourceRecordSetsRequest>}
           request({
             :body       => body,
+            :idempotent => true,
             :parser     => Fog::Parsers::DNS::AWS::ChangeResourceRecordSets.new,
             :expects    => 200,
             :method     => 'POST',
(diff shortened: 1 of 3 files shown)


geemus:
Thanks!

lib/fog/aws/requests/dns/change_resource_record_sets.rb
```diff
@@ -141,6 +141,7 @@ def change_resource_record_sets(zone_id, change_batch, options = {})
           body = %Q{<?xml version="1.0" encoding="UTF-8"?><ChangeResourceRecordSetsRequest xmlns="https://route53.amazonaws.com/doc/#{@version}/">#{changes}</ChangeResourceRecordSetsRequest>}
           request({
             :body       => body,
+            :idempotent => true,
```
lanej: is this ok for retries despite being a `POST` ?

geemus: I wondered the same thing. Usually I would say no (post generally creates things). So far as I can tell this is more like a put (updating the record) and it just replaces values. So I think you could re-replace with the same values as often or as many times as you might want without negative impact. Definitely let me know if I misunderstood though.

josacar: As it should be performed atomically, this must be idempotent as [doc](http://docs.aws.amazon.com/Route53/latest/APIReference/api-change-resource-record-sets-example-basic.html) says.

geemus: @josacar thanks for the further clarification.

lanej: @josacar thanks

PR #43, opened 2015-01-27
add 'volumeType' and 'encrypted' to blockDeviceMapping parser

ichii386:
- volumeType: standard, gp2 or io1
- encrypted: true or false


geemus:
Thanks!

PR #45, opened 2015-01-29
get signin token for federation

ehowe:


geemus:
Overall looks alright, but not functionality I'm particularly familiar with. Could you provide some context to help me understand how this fits in to things? Thanks!

lanej:
@geemus this one is a treat.

http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html

This is feature is used to bounce a user off a non-AWS identity service and into a particular IAM role without sharing the IAM credentials.

geemus:
Thanks! I think maybe this would live in fog/sts instead of in a new federation namespace though? What do you think?

ehowe:
I originally tried that, but it uses a substantially different request/response structure than most of the other endpoints use (thanks Amazon), which is why it ended up being its own namespace.  I didn't exactly want to do it this way, but it was the cleanest way to get it implemented.

geemus:
Thanks!

PR #46, opened 2015-02-01
Fix v4 signature when path has repeated slashes in the middle

fcheung:
This fixes fog/fog#3416 for me


geemus:
Thanks!

crazymykl:
Collapsing extra slashes at the front still breaks things.

PR #49, opened 2015-02-06
don't allow sg authorization is unknown sgs

lanej:
* and refactor the authorize_security_group_ingress method a bit


geemus:
Thanks!

PR #50, opened 2015-02-07
Support customer encryption headers in multipart uploads

lautis:
S3 server-side encryption with customer provided encryption doesn't work for multipart uploads since each request must contain encryption key header. This PR includes a configuration option for customer encryption key and includes necessary headers for all multipart requests.

I didn't quite get the test setup running, but manually uploading encrypted files and decrypting them seems to work. Hopefully my attempt to write tests work.


lanej:
Thanks @lautis

geemus:
Thanks!

Issue #47, opened 2015-02-03
head_url signed


mrloop:
I'm using presigned head requests, currently using [signed_url](https://github.com/fog/fog-aws/blob/b5c39ecd093da3917edea6ce7ab88f228649076d/lib/fog/aws/storage.rb#L140) but would like to use a new method on files **head_url** , similar to [get_url](https://github.com/fog/fog-aws/blob/b5c39ecd093da3917edea6ce7ab88f228649076d/lib/fog/aws/models/storage/files.rb#L80). Would a pull request with this be welcome?

geemus:
Seems like a reasonable addition to me, thanks!

mrloop:
See pull request https://github.com/fog/fog-aws/pull/51

geemus:
Thanks!

PR #51, opened 2015-02-07
Making it easier to get pre-signed head requests

mrloop:
Very similar to get_url. https://github.com/fog/fog-aws/issues/47

### DIFF
--- a/lib/fog/aws/requests/storage/head_object_url.rb
+++ b/lib/fog/aws/requests/storage/head_object_url.rb
@@ -1,0 +1,40 @@
+module Fog
+  module Storage
+    class AWS
+      module HeadObjectUrl
+        def head_object_url(bucket_name, object_name, expires, options = {})
+          unless bucket_name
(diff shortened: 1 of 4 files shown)


lib/fog/aws/requests/storage/head_object_url.rb
```diff
@@ -0,0 +1,40 @@
+module Fog
+  module Storage
+    class AWS
+      module HeadObjectUrl
+        def head_object_url(bucket_name, object_name, expires, options = {})
```
mrloop: Wondering wither this and duplicate code in get_object_url should be put in module?

geemus: Possibly, but it is a small enough amount of code I'm not too concerned about it for now anyway.

geemus:
Thanks!

mrloop:
Great, hopefully it'll be useful to somebody else.

PR #53, opened 2015-02-13
raise when destroying an ec2 firewall authorized to an rds firewall

lanej:
* get a `InUse` error when doing so for realsies


lanej:
i did 'un-pending' the tests, and they passed

See #16

geemus:
Thanks!

PR #54, opened 2015-02-17
not allowed to delete a "revoking" rds firewall

lanej:


lanej:
the mock forces you to `reload` the rds firewall in order to progress the lifecycle, which is the proper way to make sure the revoked rule is actually gone.

geemus:
Thanks!

PR #57, opened 2015-02-20
[dns] fix Records#get, mock records and proper errors

lanej:


geemus:
@lanej Looks like this breaks some tests? Could you take a look? Thanks!

lanej:
@geemus i started to fix one thing and it cascaded a little but this should be gtm.

geemus:
Thanks!

PR #58, opened 2015-02-20
Fix for ScanFilter parameters

nawaidshamim:
'AttributeValueList' needs to be an array of 'AttributeValue' and not Hash.


geemus:
Thanks!

PR #59, opened 2015-02-20
Update documentation comments for key

nawaidshamim:
Both HashKey and RangeKey should be Hash - formated as {type => value}

### DIFF
--- a/lib/fog/aws/requests/dynamodb/delete_item.rb
+++ b/lib/fog/aws/requests/dynamodb/delete_item.rb
@@ -8,11 +8,9 @@ module Fog
         # * 'table_name'<~String> - name of table for item
         # * 'key'<~Hash>:
         #   * 'HashKeyElement'<~Hash>: info for primary key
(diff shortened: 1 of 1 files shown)


geemus:
I thought in other usage I had done it ended up looking something like:

```
'HashKeyElement'  => { 'AttributeName' => 'user',  'AttributeType' => 'S' },
'RangeKeyElement' => { 'AttributeName' => 'token', 'AttributeType' => 'S' }
```

I think this pull suggests something more like:

```
'HashKeyElement'  => { 'AttributeValue' => { 'S' => 'user' } },
'RangeKeyElement' => { 'AttributeValue' => { 'S' => 'token' } }
```

Maybe this is just a newer variation (when we wrote this originally I think you had to do it the former way). Does that make sense/match up with your expectation? This newer ways seems like it might be simpler to use though, but I guess I would prefer the docs to be consistent throughout if that is the case. I think update_item at least also uses the other style. What do you think?

lanej:
Stale. Closing.

PR #60, opened 2015-02-23
Support for IAM managed policies

fcheung:
See https://aws.amazon.com/blogs/aws/new-managed-policies-for-aws-identity-access-management/


geemus:
Thanks!

PR #61, opened 2015-02-24
Add support for cname buckets

dsgh:
Ported from fog/fog#3140.

This allows users to pass `bucket_cname: true` and have the url be generated as per http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingCustomURLs

If this is missing tests, documentation, please let me know where and how it should be done. I couldn't find out much from the contributing guidelines. Thanks

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -256,21 +256,25 @@ module Fog
           if params[:bucket_name]
             bucket_name = params[:bucket_name]
 
-            path_style = params.fetch(:path_style, @path_style)
-            if !path_style
-              if COMPLIANT_BUCKET_NAMES !~ bucket_name
-                Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) is not a valid dns name, which will negatively impact performance.  For details see: http://docs.amazonwebservices.com/AmazonS3/latest/dev/BucketRestrictions.html")
-                path_style = true
-              elsif scheme == 'https' && !path_style && bucket_name =~ /\./
-                Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) contains a '.' so is not accessible over https as a virtual hosted bucket, which will negatively impact performance.  For details see: http://docs.amazonwebservices.com/AmazonS3/latest/dev/BucketRestrictions.html")
-                path_style = true
-              end
-            end
-
-            if path_style
-              path = bucket_to_path bucket_name, path
+            if params[:bucket_cname]
+              host = bucket_name
             else
-              host = [bucket_name, host].join('.')
+              path_style = params.fetch(:path_style, @path_style)
+              if !path_style
+                if COMPLIANT_BUCKET_NAMES !~ bucket_name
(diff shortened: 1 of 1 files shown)


lib/fog/aws/storage.rb
```diff
@@ -256,21 +256,25 @@ def request_params(params)
...
-            path_style = params.fetch(:path_style, @path_style)
-            if !path_style
-              if COMPLIANT_BUCKET_NAMES !~ bucket_name
-                Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) is not a valid dns name, which will negatively impact performance.  For details see: http://docs.amazonwebservices.com/AmazonS3/latest/dev/BucketRestrictions.html")
-                path_style = true
-              elsif scheme == 'https' && !path_style && bucket_name =~ /\./
-                Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) contains a '.' so is not accessible over https as a virtual hosted bucket, which will negatively impact performance.  For details see: http://docs.amazonwebservices.com/AmazonS3/latest/dev/BucketRestrictions.html")
-                path_style = true
+            if params[:bucket_cname]
+              host = bucket_name
```
geemus: I wouldn't expect bucket_name to be a full host value, but rather just the name of the particular bucket. Does that not match up with you experience/expectation?

dsgh: @geemus I'm using a bucket_name such as: `assets.example.com`. As per this [S3 documentation page](http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingCustomURLs), you can then have functioning urls in the form of: `https://assets.example.com/the/path/to.png`

I have only been able to get an url in that form by using this patch and passing in this option, but glad to be shown other ways of achieving this.

geemus: Ah. I think maybe you can do it via setting `assets` as the bucket_name and `example.com` as the host, but I'm not very certain of that.

dsgh: That would entail defining the bucket_name as `assets` on S3, and then the CNAME redirect wouldn't work.

geemus: Bah, sorry. I think maybe I'm just totally mis-remembering. Have you tried using the `endpoint` param? I think maybe setting that to your CNAME might be the right thing. Sorry for the run around here, it's been ages since I interacted with this directly.

dsgh: Ok I've tried setting the `endpoint` param and it seems a step in the right direction, but then I get an url like `https://assets.example.com/assets.example.com/the/path/to.png` because of https://github.com/fog/fog-aws/blob/master/lib/fog/aws/storage.rb#L271. Also setting `:path_style` to `false` doesn't help, as a) it's re-set to `true` in line 268; and b) even if it weren't, it means that https://github.com/fog/fog-aws/blob/master/lib/fog/aws/storage.rb#L273 would be run instead, and I would get an url like `https://assets.example.com.assets.example.com/the/path/to.png` (which results in an excon exception since that domain is not an S3 domain)

geemus:
Sorry for the delay, thanks!

PR #63, opened 2015-03-03
Mock method for AWS S3 post_object_hidden_fields 

byterussian:
Hi,

` post_object_hidden_fields ` doesn't make any external call, can be used as Mock method. Tell me if there a more elegant way to do this instead of copy & paste.

### DIFF
--- a/lib/fog/aws/requests/storage/post_object_hidden_fields.rb
+++ b/lib/fog/aws/requests/storage/post_object_hidden_fields.rb
@@ -1,7 +1,7 @@
 module Fog
   module Storage
     class AWS
-      class Real
+      module PostObjectHiddenFields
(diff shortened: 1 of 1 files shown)


geemus:
I think in other (similar) cases we have created a little module that has the method in this file. Then you can just "include" that module in to both real/mock and you don't have to copy paste. Does that make sense/work for you?

byterussian:
Yep,

I've just made in this manner. My only doubt is about the doc comment... it goes before the method declaration or inside the `Real` class?

geemus:
Hmm. I guess with the definition probably...

geemus:
Yeah, how you did it seems fine. Thanks!

PR #67, opened 2015-03-08
Stringify keys for query parameters

jfmyers9:
* The query hash expects that all keys are strings. If the query hash
  has both string and symbol keys the `signature_v2` method raises an
  exception
* The query hash is often the leftover keys that have not been
  extracted from the initial request.
* The conversion of sym -> string is done elsewhere for other parameters
* Makes the request more robust in general.


geemus:
Thanks!

PR #70, opened 2015-03-12
eu-central missing from Fog::Compute::AWS::Mock

wyhaines:
Added a couple lines so that the Mock class knows about eu-central-1a and eu-central-1b.


geemus:
Thanks!

PR #72, opened 2015-03-13
Resource record sets bug fix + support eu-central-1 

mattheworiordan:
Fixes a an error `undefined method `last' for {"eu-west-1"=>"Z3NF1Z3NOM5OY2"}:Hash` when using the `change_resource_record_sets` method.

Also added support for the new eu-central-1 region


lanej:
Thanks @mattheworiordan

geemus:
Thanks!

PR #73, opened 2015-03-13
Fix for `undefined method `map' for nil:NilClass`

mattheworiordan:
1e799c59187702650253000f80bed67e422acc5f expects that params[:query] is always defined, and also introduced another bug that reassigned params incorrectly


geemus:
Thanks!

PR #74, opened 2015-03-19
[rds] prevent final snapshot on replicas

lanej:
* `InvalidParameterCombination => FinalDBSnapshotIdentifier can not be specified when deleting a replica instance`


geemus:
Thanks!

PR #75, opened 2015-03-21
Fix blank content-encoding when none is supplied

fcheung:
Should help with fog/fog#3450

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -561,9 +561,12 @@ module Fog
               params[:headers]['x-amz-content-sha256'] = 'STREAMING-AWS4-HMAC-SHA256-PAYLOAD'
               params[:headers]['x-amz-decoded-content-length'] = params[:headers].delete 'Content-Length'
 
-              encoding = "aws-chunked"
+              if params[:headers]['Content-Encoding']
+                encoding = "aws-chunked, #{params[:headers]['Content-Encoding']}"
+              else
+                encoding = "aws-chunked, identity"
+              end
 
-              encoding += ", #{params[:headers]['Content-Encoding']}" if params[:headers]['Content-Encoding']
               params[:headers]['Content-Encoding']  = encoding
             else
               params[:headers]['x-amz-content-sha256'] ||= Digest::SHA256.hexdigest(params[:body] || '')


lautis:
With this the Content-Encoding header is set as

```
Content-Encoding:  identity
```

Note that there is two spaces between colon and identity. I'm not sure if that could have any effect.

fcheung:
Not according to http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html

geemus:
@fcheung - seems good. Have you confirmed that it helps on #3450? Figure knowing that will make it clearer if we should pull it in or not. Thanks!

fcheung:
@geemus waiting for feedback

geemus:
@fcheung Thanks, that was my impression, just double checking.

fcheung:
@geemus I think we're good to go on this

geemus:
@fcheung thanks, bringing it in.

jonbuda:
Just to follow up on this, it appears as though 'Identity' should not be used for 'Content-Encoding', according to: http://tools.ietf.org/html/rfc2616#section-3.5

Thoughts? I'm not entirely what effect it has on things though, if it is present.

wizjo:
We are experiencing a problem with this fix -- all the gzip files ended up with an empty `Content-Encoding` file header in S3. As a result, our CDN service couldn't determine the actual `Content-Encoding` and is serving assets in their original size. We reverted the change in our fork of gem "fog-aws" and are using that fork, but are there any plans to fix this issue down the road?

fcheung:
So you suffer from the opposite problem than the person that reported the original issue? If identity is an incorrect content encoding (which it seems it is) then we need to figure out what was causing the blank content encoding previously. I suppose it could happen if the content-encoding was explicitly set to an empty string, but not sure how that would happen

wizjo:
@fcheung , this particular fix: https://github.com/fog/fog-aws/blob/9b9faf6818ab27a335cbc0b9675a654ef80124c9/lib/fog/aws/storage.rb#L567 results in a blank "Content-Encoding" file header in S3, which is causing the CDN problem described above.
screen shot 2015-05-06 at 4 20 13 pm

fcheung:
I've created a pull request that changes it back and improves some of the handling, but the original fog issue was by someone having this issue *before* we made this change, so I am not confident this won't introduce a regression

geemus:
I've merged the change, could people who have experienced issues (or are now) try with master and see if it improves their situations? Thanks!

PR #78, opened 2015-03-25
Remove assignment within conditional in File#body

greysteil:
Came across this today. Hopefully this change makes the method a little easier to read.


geemus:
Seems like a good clarifying change. We tend to avoid ternaries in the style we use though, could you expand those into regular if/else statements? With that done I'll be happy to bring this in. Thanks!

greysteil:
Done!

geemus:
Thanks, looks solid!

PR #85, opened 2015-04-10
Move more requires to autoload

plribeiro3000:
On the last PR i left some requires that could been using `autoload` to improve the performance over the issue with fog/fog#3430.

### DIFF
--- a/lib/fog/aws.rb
+++ b/lib/fog/aws.rb
@@ -3,9 +3,6 @@ require 'fog/xml'
 require 'fog/json'
 
 require File.expand_path('../aws/version', __FILE__)
-require File.expand_path('../aws/credential_fetcher', __FILE__)
-require File.expand_path('../aws/region_methods', __FILE__)
-require File.expand_path('../aws/signaturev4', __FILE__)
 
 module Fog
   module CDN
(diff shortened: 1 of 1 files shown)


geemus:
Cool, thanks! Do you think this will help with the route53 cookbook issue?

plribeiro3000:
@geemus Perhaps. But i fear the issue is related to the missing require for `fog/core` in `lib/fog/aws/storage` after my changes. If the latest version of `route53` still have this issue we can just put back the require and warning all libs that depend on fog of this ongoing change in `fog`.

Before the extraction the require did make some sense because `fog` has a lazy load mechanism and the dependent libs could do a require for only the service they want to reduce memory usage.
Since we started moving to gemnized providers, this is not valid anymore because the user can just list the providers they want instead of fog and just the service they use will be loaded because of `autoload`.

So i believe this is an issue because of they way `fog` used to work and the new way it is heading.
Perhaps we might have to plain it a little more and take this kind of stuff in account before moving.

For now i think that we should fix the issue and try to not follow the old style again since it does not fit the new architecture. But perhaps we can do it for a while until we get everything working as expected.

If we don't get a response/feedback related to the `route53` version being used i will add the require back as a temporary workaround.

geemus:
Got it, thanks for the update!

PR #89, opened 2015-04-16
Support weight round robin mock

freddy1666:


lib/fog/aws/requests/dns/list_resource_record_sets.rb
```diff
@@ -59,6 +59,20 @@ def list_resource_record_sets(zone_id, options = {})
       end
 
       class Mock
+        def set_identified_records(record, zone, name)
```
geemus: I feel like the `set` here is potentially confusing. I imagine it to mean like "a set of objects", but I can't help but read it like "set a value". I feel like it might be clearer if it were just `get_identified_records`. What do you think?

geemus:
Looks good overall, added one comment above. Also it looks like there are some failures on travis for the tests relating to `undefined method `append' for []:Array (NoMethodError)`. If you could also check in on those that would be great. Thanks!

freddy1666:
the bug should be fixed

geemus:
Thanks!

PR #90, opened 2015-04-16
adding support for d2 instance type

yumminhuang:
AWS has released [Dense-storage Instances](https://aws.amazon.com/blogs/aws/next-generation-of-dense-storage-instances-for-ec2/) for EC2.

Modify ` lib/fog/aws/models/compute/flavors.rb` for adding d2 instance types.


geemus:
Thanks!

PR #91, opened 2015-04-17
Fix repository URL in README.md

tricknotes:


geemus:
Oops, thanks for catching our slip up!

PR #92, opened 2015-04-17
mocks for sns topic permissions

ehowe:


geemus:
Overall it looks good, but it seems to break in the 1.8 related tests, could you take a look and see if you can pin it down? I think perhaps because select on a hash in 1.8 returns an array or something?

geemus:
P.S. same thing came up elsewhere, so maybe check #94 for a possible fix.

PR #94, opened 2015-04-17
Fix mock VPC ELB creation in regions other than us-east-1

mrpoundsign:
When mocking, the way it was getting the region would 
always fail if you weren't passing in availability zones. 
Fixed so it finds the region for subnet_ids passed in.

### DIFF
--- a/lib/fog/aws/requests/elb/create_load_balancer.rb
+++ b/lib/fog/aws/requests/elb/create_load_balancer.rb
@@ -74,9 +74,19 @@ module Fog
           dns_name = Fog::AWS::ELB::Mock.dns_name(lb_name, @region)
 
           availability_zones = [*availability_zones].compact
-          region = availability_zones.empty? ? "us-east-1" : availability_zones.first.gsub(/[a-z]$/, '')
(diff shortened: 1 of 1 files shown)


geemus:
This looks good, except unfortunately not working in 1.8 (which we are, at least for now still trying to support).  I think select on hash in 1.8 returns an array instead of a hash, which breaks expectations here, could you take a look at travis and see if you can work something out? Thanks!

mrpoundsign:
@geemus yeah I saw that. Trying to figure it out now.

geemus:
Thanks!

mrpoundsign:
@geemus it's kind of dirty, but I think I can Hash[<the select statement>] -- Seems to work in 2.0 also. Any better suggestions?

    2.0.0-p247 :003 > a = [["b", 200], ["c", 300]]
    2.0.0-p247 :004 > Hash[a]
    {
        "b" => 200,
        "c" => 300
    }
    2.0.0-p247 :005 > Hash[Hash[a]]
    {
        "b" => 200,
        "c" => 300
    }

mrpoundsign:
Still waiting for the final build but it looks like it's passing now :)

geemus:
Cool, yeah, sort of weird, but works-for-me. Thanks!

PR #96, opened 2015-04-26
Parse elasticache configuration endpoint from response

fcheung:
retrieves the configuration endpoint from the response (as used by the auto discovery feature for memcached elasticache clusters)


geemus:
Thanks!

PR #97, opened 2015-04-27
Adding instanceTenancy to reserved instance parser.

dmbrooking:
Simple change adding one field to the reserved instance parser. Field is present on response per: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeReservedInstances.html


geemus:
Thanks!

Issue #100, opened 2015-04-29
IAM authentication not compatible with GovCloud 


benbalter:
In https://github.com/fog/fog-aws/blob/master/lib/fog/aws/iam.rb#L233, the IAM implementation assumes the region us `us-east-1` for all requets. 

```ruby
#global services that have no region are signed with the us-east-1 region
@signer = Fog::AWS::SignatureV4.new( @aws_access_key_id, @aws_secret_access_key,'us-east-1','iam')
```

This is not true for AWS [GovCloud](https://aws.amazon.com/govcloud-us/), the government-specific AWS region.

If the host is set as `iam.us-gov.amazonaws.com`, a request for `Fog::AWS::IAM#roles` will return:

`Credential should be scoped to a valid region, not 'us-east-1'.`

Instead, I'd suggest that the IAM constructor accept a region argument (to mirror other constructors), which defaults to `us-east-1`, to allow for better GovCloud support.

geemus:
@benbalter sounds reasonable, I suspect we just hadn't stumbled upon this just yet. Would you be up for working on a pull request to fix this? Thanks!

benbalter:
> Would you be up for working on a pull request to fix this?

Sure thing. Done over in https://github.com/fog/fog-aws/pull/102.

geemus:
Awesome, thanks!

PR #102, opened 2015-05-01
Allow the IAM constructor to accept a region

benbalter:
This should improve support for GovCloud, which requires requests to be scoped to its own region.

Fixes https://github.com/fog/fog-aws/issues/100.


geemus:
Thanks!

PR #103, opened 2015-05-02
Refresh credentials if needed when signing S3 URL

matkam:
Should fix https://github.com/fog/fog/issues/3544, https://github.com/carrierwaveuploader/carrierwave/issues/1547


geemus:
Thanks!

PR #105, opened 2015-05-06
ami: Show CreationDate of AMIs.

amerine:
Hey hey, CreationDate has been available on AMIs for some time. I believe this will expose it to `Fog::Compute::AWS::Image` objects.

### DIFF
--- a/lib/fog/aws/models/compute/image.rb
+++ b/lib/fog/aws/models/compute/image.rb
@@ -22,6 +22,7 @@ module Fog
         attribute :tags,                  :aliases => 'tagSet'
         attribute :name
         attribute :virtualization_type,   :aliases => 'virtualizationType'
+        attribute :creation_date,         :aliases => 'creationDate'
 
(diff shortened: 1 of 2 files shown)


geemus:
Looks like a good start, some extra bits that we should probably tack next:

* I think you will also need to add it to the parser to get it back from AWS here: https://github.com/amerine/fog-aws/blob/ami-creation-date/lib/fog/aws/parsers/compute/describe_images.rb

* You'll also want to add it to the mock for register images, so that it will have a value: https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/compute/register_image.rb#L69

* Finally, should probably add some tests to make sure it is doing what we think and prevent regressions.

Let me know if you have any questions or anything I can help with. Thanks!

estonfer:
covered by https://github.com/fog/fog-aws/pull/321

geemus:
@estonfer good catch, thanks!

PR #106, opened 2015-05-07
Another attempt to solve content-encoding header issues

fcheung:
see #75


geemus:
Thanks!

Maxim-Filimonov:
I don't think it's fixed.
We are experiencing empty content-type with v0.4.0 when using with middleman-s3_sync needed to downgrade all the way to 0.1.2 to fix it.

PR #113, opened 2015-05-16
Fix v4 signature when path has leading or trailing repeated slashes

sandbox:
This adds on to fog/fog-aws#46 and fixes more cases of fog/fog#3416 for me.

Feels silly but S3 keys with leading or trailing slashes are valid when creating an S3 item. Previously in Fog 1.15.0 I was able to create files with keys like `/test//save.txt`.

Since upgrading to 1.30.0 with fog-aws 0.2.2, I am seeing the following error:
```
> connection = Fog::Storage.new(FOG_CONFIG)
> storage = connection.directories.get(aws['s3']['bucket_name'])
> storage.files.create(key: "/test//save.txt", body: "this is the body")

Excon::Errors::Forbidden: Expected(200) <=> Actual(403 Forbidden)
excon.error.response
  :body          => "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>SignatureDoesNotMatch</Code>
<Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
...</Error>"
  :headers       => {
    "Connection"       => "close"
    "Content-Type"     => "application/xml"
    "Date"             => "Sat, 16 May 2015 21:08:25 GMT"
    "Server"           => "AmazonS3"
    "x-amz-id-2"       => "xxx"
    "x-amz-request-id" => "xxx"
  }
  :local_address => "x.x.x.x"
  :local_port    => 56051
  :reason_phrase => "Forbidden"
  :remote_ip     => "x.x.x.x"
  :status        => 403
  :status_line   => "HTTP/1.1 403 Forbidden\r\n"
```

This pull request allows repeated leading and trailing slashes.

### DIFF
--- a/lib/fog/aws/signaturev4.rb
+++ b/lib/fog/aws/signaturev4.rb
@@ -79,8 +79,6 @@ DATA
       protected
 
       def canonical_path(path)
-        #leading and trailing repeated slashes are collapsed, but not ones that appear elsewhere
-        path = path.gsub(%r{\A/+},'/').gsub(%r{/+\z},'/')
         components = path.split('/',-1)
         path = components.inject([]) do |acc, component|
           case component
--- a/tests/signaturev4_tests.rb
+++ b/tests/signaturev4_tests.rb
@@ -58,7 +58,7 @@ Shindo.tests('AWS | signaturev4', ['aws']) do
 
   tests('get with repeated / ') do
     returns(@signer.sign({:query => {}, :headers => {'Host' => 'host.foo.com', 'Date' => 'Mon, 09 Sep 2011 23:36:00 GMT'}, :method => :get, :path => '//'}, @now)) do
-      'AWS4-HMAC-SHA256 Credential=AKIDEXAMPLE/20110909/us-east-1/host/aws4_request, SignedHeaders=date;host, Signature=b27ccfbfa7df52a200ff74193ca6e32d4b48b8856fab7ebf1c595d0670a7e470'
+      'AWS4-HMAC-SHA256 Credential=AKIDEXAMPLE/20110909/us-east-1/host/aws4_request, SignedHeaders=date;host, Signature=ed90a993076c9d2289f0942851cefcca04936604d925e4e3d627a046d308efbe'
     end
   end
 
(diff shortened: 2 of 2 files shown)


lanej:
it would appear the permission of sequential forward slashes might be contextual. 

paging @fcheung as he worked on https://github.com/fog/fog-aws/pull/46

fcheung:
Hmm, the test cases this commit changes are both from the AWS test suite (http://docs.aws.amazon.com/general/latest/gr/signature-v4-test-suite.html) - I would be very reticent to change those tests.

We don't have a testcase for /test//save.txt (not can I see one in the amazon docs) so we should fix that - I will take a look

geemus:
Thanks!

sandbox:
Thanks for taking a look, would it help if I added some of the tests from the AWS test suite? I wrote them up but the build will fail, does that matter? I'm relatively new to contributing

Interestingly, I get the same SignatureDoesNotMatch error when attempting `storage.files.create(key: "//foo//", body: 'body')`, maybe there is an option to specify to get the correct signature when performing the request? it succeeds with the pull request's change but the signature test still has a different result than the AWS example.

fcheung:
More tests is a good thing! Hopefully they shouldn't break the build, unless you uncover more issues (a complete pipe dream would be the ability to just use the sample data in the test suite download, rather than having to translate each one into ruby). However I've spotted that there is a discrepancy between the signature v4 docs and the s3 docs - it would seem that s3 is different from all the other services, and the implementation in the aws ruby sdk backs this up (I am pursuing this further)

Fred

fcheung:
I've opened a support case with AWS to get them to look at the discrepancy between the s3 docs and the signature docs

fcheung:
Finally heard back from them. S3 implements signature v4 different from every other service, in a way that doesn't conform to the test cases (namely it doesn't do path canonicalisation).

geemus:
@fcheung thanks for chasing that down. So, I guess we need a distinct signer? Or at least a cononacalization parameter of some sort.

fcheung:
Yeah, it's pretty straightforward, since it's only the path canonicalization that changes and the signer already knows what service it is signing for. Will try to throw something together in the next day or 2.

geemus:
Sounds good, thanks!

PR #114, opened 2015-05-18
support iam/get_user without username

lanej:
* returns current user


lanej:
passes for realsies

geemus:
Thanks!

PR #117, opened 2015-05-20
Update that will allow updating global indexes

logicwreck:
Update that will allow updating global indexes.
 def update_table(table_name, provisioned_throughput)
changed to
def update_table(table_name, provisioned_throughput, global_indexes=nil)

where global_indexes should be a hash specified like at http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateTable.html

### DIFF
--- a/lib/fog/aws/requests/dynamodb/update_table.rb
+++ b/lib/fog/aws/requests/dynamodb/update_table.rb
@@ -27,7 +27,7 @@ module Fog
         #       * 'TableName'<~String> - name of table
         #       * 'TableStatus'<~String> - status of table
         def update_table(table_name, provisioned_throughput, global_indexes=nil)
-          if global_indexes == nil
+          if global_indexes != nil
            body = {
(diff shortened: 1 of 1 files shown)


geemus:
Seems like a great addition. Maybe instead of a single extra, optional argument, we should rework it toward an second argument as an options hash? We can check and remap provisioned_output into the hash with a deprecation warning or just allow through a hash. I think the benefit would be that it might be more future-proof this way (we have some examples of this already in the codebase). What do you think?

logicwreck:
I think this is also a good idea. The main idea of this addition was so that it fully updates the dynamoDB table which was not happening with the existing version of the function. I needed this for a task thus I did this modification.

geemus:
Cool, yeah that was definitely a reasonable place to come from. Would you be up for making the adjustment we discussed and then we can bring this in? There should be some other examples in EC2/compute if you need them, and certainly happy to answer any questions/help if you need anything. Thanks!

lanej:
Stale. Closing.

PR #119, opened 2015-05-20
fix root iam arn

lanej:
* mock only change
* root user arn doesn't have 'user/' key prefix


lanej:
Relevant commits are part of #121

PR #123, opened 2015-05-22
Add support for AWS Lambda service

nomadium:
This is obviously not complete, I just implemented one action from AWS Lambda API since I want to discuss and consult with maintainers what would be the best approach for this service implementation.


nomadium:
AWS Lambda API is REST and content-type responses are application/json. I was checking existing fog codebase for examples but I was unable to find one so I guess this would be the first service needing to parse JSON responses from API.

```Fog::XML::Connection``` would not be appropriate to use since it expects an SAX parser for XML responses, right?

What do you think about this? Does a class Fog::JSON::Connection should be implemented? If so, what would be a good JSON parser to build upon? JSON::Stream[1] from @dgraham?

Thanks,

1. https://github.com/dgraham/json-stream

geemus:
fog/fog has some json, particularly for openstack and related (rackspace). I think you should find some good precedents there to work from, but just let me know if you have further questions. Thanks!

nomadium:
Great, I'll take a look at those.

nomadium:
This is not ready yet but I already implemented all CRUD API actions for Lambda functions. I guess during this week or next weekend I'll complete this work.

geemus:
Sounds great, just let me know if there is anything I can do to help!

nomadium:
@geemus This PR is ready for review. Thanks.

geemus:
Awesome, thanks!

PR #128, opened 2015-06-01
make elb/policies collection standalone

lanej:
* there is no need to inject data from the load balancer model when
  an index request exists that maps directly

### DIFF
--- a/lib/fog/aws/models/elb/policy.rb
+++ b/lib/fog/aws/models/elb/policy.rb
@@ -8,12 +8,13 @@ module Fog
         attribute :expiration, :aliases => 'CookieExpirationPeriod'
         attribute :type_name
         attribute :policy_attributes
+        attribute :load_balancer_id
 
         attr_accessor :cookie_stickiness # Either :app or :lb
 
         def save
-          requires :id, :load_balancer
-          args = [load_balancer.id, id]
+          requires :id, :load_balancer_id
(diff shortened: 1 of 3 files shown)


lib/fog/aws/models/elb/policy.rb
```diff
@@ -45,12 +44,8 @@ def destroy
           reload
         end
 
-        def reload
-          load_balancer.reload
-        end
-
         def load_balancer
-          collection.load_balancer
+          service.load_balancers.new(identity: collection.load_balancer_id)
```
geemus: Could you fix the 1.8 incompatible named argument here? Hopefully we'll reach a point where we can drop 1.8 soon, but not quite ready for the major/breaking change I'm afraid. Thanks!

lanej: gets me everytime.

geemus: Yeah, stuff like this is starting to bite us more and more often. Breaking changes just also are, well, breaking, so caution prevails. Would love to cut it out soon though.

lanej: it's the last "," before the end of a Hash definition that is perhaps the biggest annoyance.  Quite possible the most understated Ruby improvement.

geemus:
Thanks!

PR #132, opened 2015-06-12
Add New M4 Instance Type

yumminhuang:
Modify `lib/fog/aws/models/compute/flavors.rb` for adding [M4 instance](https://aws.amazon.com/blogs/aws/the-new-m4-instance-type-bonus-price-reduction-on-m3-c4/) flavors.


geemus:
Thanks!

yumminhuang:
@fcheung According to [Instance Types Matrix](https://aws.amazon.com/ec2/instance-types/), I think you are right. `:ebs_optimized_available` should be true for m4 instances.

PR #133, opened 2015-06-12
Fix S3 signature v4 signing

fcheung:
S3 implements signature v4 different from other services - it doesn't canonicalize the path

With this pull requests I can store keys such as `//fred//..//tes!$%&t`

See also #113


geemus:
Thanks!

PR #135, opened 2015-06-17
Make Mock create_vpc method arity match Real

fdr:
This difference in method arity was spuriouly breaking tests that have
nothing to do with those options being passed (InstanceTenancy in my
case).


lanej:
good find.

geemus:
Thanks!

PR #139, opened 2015-06-24
list_hosted_zones expects that options to be hash with symbol as key

slashmili:
If the hash key is string it won't get applied in list_hosted_zones.


geemus:
Good catch, thanks!

PR #140, opened 2015-06-25
Escape Lambda function name in request paths

nomadium:
I noticed a bug in some AWS Lambda requests. That API allows you to invoke functions by specifying the function id in several ways:

1. plain_function_name
2. account_id:function_name
3. arn:aws:lambda:region:account_id:function:function_name

I noticed I forgot to escape function names for cases 2 and 3. This PR fixes that.

Thanks,


geemus:
Yeah, if only AWS didn't have so many weird exceptions like this. Thanks for the quick fix!

PR #142, opened 2015-06-28
The :geo_location attribute needs to be xml formatted before calling aws

carloslima:
Without this, calling `record.destroy` on a record that has geolocation set will cause an exception with:

```
Fog::DNS::AWS::Error: MalformedInput => Unexpected complex element termination
```

### DIFF
--- a/lib/fog/aws/requests/dns/change_resource_record_sets.rb
+++ b/lib/fog/aws/requests/dns/change_resource_record_sets.rb
@@ -61,84 +61,7 @@ module Fog
         #     change_resource_record_sets("ABCDEFGHIJKLMN", change_batch_options)
         #
         def change_resource_record_sets(zone_id, change_batch, options = {})
-          # AWS methods return zone_ids that looks like '/hostedzone/id'.  Let the caller either use
-          # that form or just the actual id (which is what this request needs)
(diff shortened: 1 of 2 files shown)


carloslima:
I moved the request body generation out of the `change_resource_record_sets` method so that I could test it separately.
I'm not very familiar on how the tests should be structured for this project; let me know if that was a mistake :-)

Is there a need to validate the attribute before turning to xml? If so, should it happen in `change_resource_record_sets_data` or somewhere else?
Or, if anything else is not good, let me know and I'd be more than happy to fix.

Cheers!

carloslima:
The diff on github's web interface looks a lot more messier than it actually is.
The no-whitespace version looks a lot more clear: https://github.com/fog/fog-aws/pull/142/files?w=1

:-)

lanej:
@carloslima non-whitespace is *much* better, thanks.

lanej:
@carloslima there doesn't appear to be a great way to test request body transformation atm.  @fog/fog-aws might have some better input.

carloslima:
Thanks! :smiley:

geemus:
Yeah, afraid we haven't really tested this explicitly in the past. Not sure how best to add it in now.

geemus:
Thanks!

PR #144, opened 2015-07-01
Adds isDefault to parser for describe_vpcs

gregburek:
Missing attribute documented in https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Vpc.html


plribeiro3000:
Thanks!

geemus:
Thanks!

PR #146, opened 2015-07-06
Describe vpcPeeringConnectionId

fdr:
Fill in this missing element of the return tuple.

Also CCing @gregburek who is interested in this patch.


lanej:
verified that `vpcPeeringConnectionId ` is part of [DescribeRouteTables](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRouteTables.html)/[RouteTable](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RouteTable.html)/[Route](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Route.html)

geemus:
Thanks!

Issue #130, opened 2015-06-05
Still empty content-encoding when it is not set


Maxim-Filimonov:
We are experiencing empty content-type with v0.4.0 when using with middleman-s3_sync needed to downgrade all the way to 0.1.2 to fix it.

rhomeister:
I can confirm this. Using:

- `asset_sync` version `1.1.0'
- `fog` version `1.30`

Thanks @Maxim-Filimonov for suggesting a (temporary) fix to this problem. Really has giving me headaches with users complaining about botched CSS and JS.

geemus:
@fcheung I know you have looked at this previously, any ideas where we might be going off the rails still? Thanks!

fcheung:
Not really. I had a look at boto and it seems to implement as we did, and the aws ruby sdk doesn't implement the streaming upload stuff at all (it looks like it would read the entire file into memory in order to compute its sha). I opened a support ticket with AWS about it but haven't heard back yet.

fcheung:
AWS have confirmed to me that they can reproduce this, ie it isn't an issue with the fog implementation of this. No workarounds for now other than avoid this code path (ie use signature v2, or don't use the stream upload feature (in fog this is triggered if the body responds to :read)). They have asked me for more details of issues caused by the blank header - can someone chime in with that (exact symptoms, browsers affected etc)

geemus:
@fcheung got it, thanks again for digging in on this.

fcheung:
@rhomeister @Maxim-Filimonov if you have details on what symptoms end users are seeing that would be helpful to AWS. In the mean time passing `aws_signature_version: 2` should fix things (unless connecting to a region that doesn't support v2)

Maxim-Filimonov:
@fcheung It affects CDN we use - Cloudflare as it cannot determine encoding when content-encoding is set to empty string. I heard about other CDN's but I don't remember exact names

rhomeister:
I'm also using Cloudflare with S3, but I'm not sure if Cloudflare is the cause of this problem. The symptoms are that the gzipped version gets served as the plain one, resulting in garbled CSS or JS. This happens in certain browsers, among which IE.

fcheung:
Thanks for the info - I have passed it on

oskarpearson:
I can also confirm this.

Is there any way to force aws_signature_version to 2 when using asset sync?

Gems:
```
Fog - 1.31.0
Fog-aws - 0.5.0
```

[...]

PR #147, opened 2015-07-08
Fix blank content-encoding headers

fcheung:
The S3 docs claim that you should set the content-encoding to aws-chunked
for a streaming upload. This however results in S3 serving files with a blank
content-encoding header. After discussion with AWS support, it appears
that setting the content-encoding to aws-chunked isn't actually necessary
since S3 can infer this from other headers present in the request

See #130, #106 , #75


lanej:
good hunting @fcheung

geemus:
Yeesh, thanks @fcheung for continuing to chase this one.

PR #149, opened 2015-07-08
Fix broken xmlns in DNS requests

decklin:
This method was broken up into two in 0.7.0, but the big one was changed to a class method. The XML body string, however, still interpolated `@version` to set the XML namespace, and ended up being `https://route53.amazonaws.com/doc//`, which causes AWS to refuse to parse the XML, returning:

    Cannot find the declaration of element 'ChangeResourceRecordSetsRequest'

I was unable to change any RRs until I downgraded to 0.6.0.

I added an `xmlns` check to the test, but it's just a hardcoded string. Maybe an enterprising person could move all the default versions for all AWS services to constants make sure that every `xmlns` is tested somewhere. (Better yet, don't build XML by concatenating strings?) But I'd prefer to just fix this at the moment.


lanej:
@decklin Thanks for the fix.

geemus:
Thanks!

PR #152, opened 2015-07-10
encrypted storage on rds

ehowe:


lib/fog/aws/models/rds/server.rb
```diff
@@ -116,27 +116,28 @@ def attributes_to_params
...
             'DBParameterGroupName'          => parameter_group_name || attributes['DBParameterGroupName'],
             'DBSecurityGroups'              => security_group_names,
-            'DBInstanceIdentifier'          => id,
-            'AvailabilityZone'              => availability_zone,
-            'DBInstanceClass'               => flavor_id,
-            'Port'                          => port || attributes['Port'],
+            'DBSubnetGroupName'             => db_subnet_group_name,
             'Engine'                        => engine,
             'EngineVersion'                 => engine_version,
+            'StorageEncrypted'              => storage_encrypted,
```
lanej: this is actually the only change here right ?

ehowe: yeah, but the formatting specs need to be fixed before it can be merged

PR #154, opened 2015-07-18
Route53 zone listing fix and support for private hosted zones

solud:


lib/fog/aws/models/dns/zones.rb
```diff
@@ -10,8 +10,8 @@ class Zones < Fog::Collection
         model Fog::DNS::AWS::Zone
 
         def all(options = {})
-          options[:marker]   ||= marker
-          options[:maxitems] ||= max_items
+          options[:marker]   ||= marker unless marker.nil?
+          options[:maxitems] ||= max_items unless max_items.nil?
```
lanej: this should probably set in the request.  there is a lot of inconsistencies around page handling.  doesn't need to be cleaned up here.

geemus:
Thanks!

PR #155, opened 2015-07-23
restore db instance fom db snapshot

michelleN:
* add mock
* add restore method on servers collection
* modify save method on server model
* add tests


geemus:
@michelleN any questions or anything I can help with around this?

PR #158, opened 2015-07-28
update #promote_read_replica mock

lanej:
* use state "modifying"
* alias #promote to #promote_read_replica
* raise correct error when attempting to promote a master server


geemus:
Thanks!

PR #161, opened 2015-08-04
storage request bad xml schema for put bucket notification fix

bahchis:
Fixes what is probably a bad copy paste in code, so the put bucket notification xml verified and constructed correctly.


geemus:
Yeah, definitely looks like a copy/paste mistake. Thanks!

PR #165, opened 2015-08-06
volume#key_id and encrypted tests

### LINKED ISSUE #141: Support for KMS ID for EBS Volume
Currently when creating an ebs volume you can set "encrypted" to true and fog will create the encrypted volume. However it uses the master EBS key on the account to do this.

It would be great if there was a "kms_key_id" option like in the AWS CLI to allow you to specify the Key to use for encryption. The value would be a string (Key ARN).

The use in AWS CLI is documented here:
http://docs.aws.amazon.com/cli/latest/reference/ec2/create-volume.html

I dont have the time or knowledge to do a PR for this at the moment so was going to post it none the less. If I get time I will come back and attempt it myself, may be a while though.

lanej:
* Fixes #141

### DIFF
--- a/lib/fog/aws/models/compute/volume.rb
+++ b/lib/fog/aws/models/compute/volume.rb
@@ -9,10 +9,11 @@ module Fog
         attribute :created_at,            :aliases => 'createTime'
         attribute :delete_on_termination, :aliases => 'deleteOnTermination'
         attribute :device
+        attribute :encrypted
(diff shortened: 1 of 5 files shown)


geemus:
Thanks!

kylegoch:
Thanks @lanej and @geemus  ! 
I feel silly for writing a workaround to this rather than fixing it here. It did not look as complex as I had made it out to be in my head.

PR #178, opened 2015-09-01
add IAM role paging

### LINKED ISSUE #176: IAM roles.all should paginate
As of 0.7.6, if you call iam.roles.all, and you have > 100 roles, Fog::AWS::IAM fails to use IsTruncated and Marker to request the next batch of roles.

I did some experimenting with this, and if I updated https://github.com/fog/fog-aws/blob/master/lib/fog/aws/models/iam/roles.rb#L13 to accept options, I could make a call like this:
iam_roles = iam.roles.all('MaxItems' => 1000)
and that would get all of my roles in one call.

Spent a couple hours poking at this and couldn't seem to find the right combo to get pagination working properly...

lanej:
* Fixes #176

### DIFF
--- a/lib/fog/aws/iam/paged_collection.rb
+++ b/lib/fog/aws/iam/paged_collection.rb
@@ -13,7 +13,9 @@ module Fog
           to_a.each(*args, &block)
         end
 
-        def each
+        def each(options={})
+          limit = options[:limit] || 100
+
           if !block_given?
             self
           else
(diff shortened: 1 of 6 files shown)


geemus:
@lanej looks like there is a hash colon instead of hash rocket, which broke the 1.8.7 builds. Could you fix? FWIW, I think fog/fog-core will be going to 2.0.0 (with the major changed being dropping 1.8.7) soon (tomorrow if things go smoothly). But should maybe still do compatibility here for the moment.

lanej:
1.8 gets me everytime.

geemus:
Thanks!

PR #182, opened 2015-09-11
Yield arguments to Mock#get_object block more similar to Excon

tdg5:
Minor tweak to Mock#get_object to make the arguments yielded to a given block more closely match the arguments that are passed by Excon in the Real case.

Tests seem to be broken on master right now :crying_cat_face:


geemus:
Thanks, tests seemed to be passing again by the time I got to this, so somebody must have addressed.

PR #184, opened 2015-09-22
mock instance stop and start properly

ehowe:
* `Server#stop` was returning the wrong state
* `#stop_instances` and `#start_instances` mocks did not include the `requestId`
* tests for all of the above


geemus:
Thanks!

PR #185, opened 2015-09-29
Treat storage-class header case-agnostically

ANorwell:
The HTTP spec specifies that headers are case-insensitive, but fog
does not treat them that way. In general, this is not a problem because
excon handles this, but when using VCR, headers are recorded as camel-case,
which are then not recognized by fog-aws.

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -20,7 +20,7 @@ module Fog
         attribute :last_modified,       :aliases => ['Last-Modified', 'LastModified']
         attribute :metadata
         attribute :owner,               :aliases => 'Owner'
-        attribute :storage_class,       :aliases => ['x-amz-storage-class', 'StorageClass']
+        attribute :storage_class,       :aliases => ['x-amz-storage-class', 'X-Amz-Storage-Class', 'StorageClass']
         attribute :encryption,          :aliases => 'x-amz-server-side-encryption'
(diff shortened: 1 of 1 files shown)


geemus:
@ANorwell looks like you closed this shortly after creation. Anything I can help with here? Thanks!

ANorwell:
@geemus This PR was meant to be against downstream (https://github.com/backupify/fog-aws/pull/1). I don't think this is ready for merging, and in fact this might either be an issue with excon or VCR, rather than fog. I haven't been able to dig into those gems yet, but any guidance would be appreciated.

The symptoms of the issue are as follows: s3 returns all lowercase headers (e.g. "x-amz-storage-class"). If the VCR gem is in use, it records these response headers with capitalization (e.g. "X-Amz-Storage-Class"). On a subsequent playback of the VCR cassette, the headers will have that same capitalization, and not be recognized by fog.

It seems that VCR is at somewhat at fault for recording the response inaccurately (possibly due to excon header normalization?). However, fog should possibly be more permissive on header case, since the HTTP spec does specify that headers are case-insensitive.

I can open an issue if that makes sense.

geemus:
Hmm. I'm not sure off hand. Maybe make sure you have the latest versions of excon/vcr. I think we made the behavior there more sane/expected than it used to be a while back, though I don't remember now if it would have been likely to specifically address this or not. Otherwise, as you indicate, it may require digging a bit more deeply.

PR #187, opened 2015-10-01
add cn support for s3

ming-relax:
s3 endpoint for `cn-north-1` is different. Tested using my account on aws china.


geemus:
Thanks!

PR #189, opened 2015-10-07
Allow region to be set for STS

fcheung:
STS is now available in all regions. (although somewhat oddly you have to manually enable each region other than us-east-1 from the AWS console - see https://blogs.aws.amazon.com/security/post/Tx3CYWU11LY2GLB/AWS-Security-Token-Service-Is-Now-Available-in-Every-AWS-Region )


geemus:
Thanks!

PR #190, opened 2015-10-13
Cache cluster security group parser

eherot:
Not sure if this is a bug or just Fog not keeping pace with the AWS API but it looks like ElastiCache security groups took on a new format and are now available under a key called `SecurityGroups`. This PR adds parsing for that key without dropping parsing for the old one. Let me know what you think.

Note that I didn't add any unit tests because there did not appear to be a test framework already set up for ElastiCache in Fog and I don't really have the cycles (or the knowledge, really) to create one just to test this one fix.


geemus:
Probably just a matter of not keeping up, but we may have missed it. Have you manually tested this? If it was us falling behind, new changes usually don't come in without bumping the API version requested, so I wanted to make sure we weren't missing that.

eherot:
Aye, I manually tested it against our existing infrastructure and it seemed to work pretty well.

geemus:
Sounds legit and at least shouldn't break anything (ie old responses should be fine, it is an additive change). Thanks!

eherot:
Welcome!

PR #197, opened 2015-11-11
make net/ssh require optional

geemus:


plribeiro3000:
:+1:

lanej:
still required by fog-core https://github.com/fog/fog-core/blob/master/fog-core.gemspec#L27.  what's the reason it's required there?

geemus:
@lanej sorry for the lack of context. I have a PR open on fog-core to make it not a requirement (basically it will try to load it only if/when you need it, and if it isn't there it errors and lets you know to add it yourself). Many users never need it, so limiting the deps is nice. Plus, net-ssh stopped supporting 1.9.3, which I don't think we are quite ready to do, so this buys us a bit more time to do that on our own terms. Just let me know if you have questions or concerns. Thanks!

lanej:
@geemus got it. judging by the failures, i think that fog-core might need to go out first so we have the correctly bounded version of net-ssh for the < ruby 2.x builds.

geemus:
Yeah, that was the impression I had as well. Trying to stage similar changes on the various repos (some don't need to go in this order, so trying to get them out of the way first).

geemus:
Hmm, travis is being weird. Looks like it passed here: https://travis-ci.org/fog/fog-aws/builds/90557817

Issue #196, opened 2015-11-06
S3 KMS encryption support


shuhei:
Is S3 server-side encryption with KMS-managed keys supported?

http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingKMSEncryption.html

It does not seem to be supported yet but I'd like to confirm before starting to work on it :-)

https://github.com/fog/fog-aws/blob/master/lib/fog/aws/storage.rb

geemus:
It's not totally clear to me. I think you can pass other things through as options on put_object for instance: https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/put_object.rb#L38. But it doesn't document the server side encryption related headers. It shouldn't hurt to try it, and I suspect it might work, but we might also need to at least bump the API version (and even if it does work, we should still improve the documentation to make this more abundantly clear).

We would certainly welcome your help pinning down the current state and making it clear (and/or making it work in the first place). Thanks!

shuhei:
Thanks for your quick response! I'm gonna try it first and report the result.

shuhei:
It worked! I use `carrierwave` and it's config is like:

```rb
CarrierWave.configure do |config|
  config.fog_credentials = {
    provider: 'AWS',
    # snip
  }
  config.fog_public = false
  config.fog_directory = ENV['AWS_S3_BUCKET']
  config.fog_attributes = {
    encryption: 'aws:kms',
    'x-amz-server-side-encryption-aws-kms-key-id': ENV['AWS_S3_ENCRYPTION_KEY_ID']
  }
}
```

Thanks for taking your time :-)

geemus:
Glad to hear it. We would certainly welcome any help around improving docs to make this clearer if you have time/inclination. Thanks!

shuhei:
Sure thing. Where is the right place to update documentation? [The code comments](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/put_object.rb#L5-L26)?

geemus:
@shuhei yeah, that would be a great place to start. I don't think we have any other place to put that kind of info at present, but I could see us working toward adding that sort of thing in time. Thanks!

PR #199, opened 2015-11-15
Document options for S3 server-side encryption

shuhei:
#196 

Document options for [the three ways of S3 server-side encryption](http://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html).


geemus:
Thanks!

PR #200, opened 2015-11-20
Add log warning message about when not on us-region

kitofr:
- Non us-regions are not supported for mocks. Added message


geemus:
The error raised here is intended to be inline with the behavior you would see in live usage (I think). Maybe it would be better to use Fog.warning and leave the error unchanged? What do you think?

kitofr:
That sounds like a fair solution. I merely wanted to get some heads up about the region problem whilst using Mocks.
I'll see if I can change it to use `Fog.warning` instead.

geemus:
Sounds good, thanks!

lanej:
@kitofr please rebase off of master

kitofr:
Well... this is over a year old. I you guys want to improve your code, please fix this. I'll leave it up to you.

geemus:
Thanks!

PR #201, opened 2015-11-20
Fix rubyzip 1.x compatibility.

voxik:


voxik:
Ah, Ruby 1.8.7 :/

lanej:
Stale.  Closing.

geemus:
Thanks.

voxik:
The patch is still valid and we still carry it around in Fedora, since we have rubyzip 1.x+. I would more appreciate your feedback then closing this issue as a stalled.

lanej:
@voxik Until we drop 1.8 support the best solution would be to modify the 1.8.7 Gemfile to use the 0.9 release of rubyzip.  You'll also need to do some version detection or class aliasing in the lib file.

PR #202, opened 2015-11-23
filters on tags can pass an array

craiggenner:
this comparison works fine where there is just a string or single tag,
where there is an array of tags this comparison breaks fails.

So instead we must cast to an array and then do a comparison by an
inclusion of a a value in an array


geemus:
Thanks!

craiggenner:
Thanks for the merge.

PR #205, opened 2015-11-30
Reset 'finished' when rewinding S3Streamer

jschneiderhan:
I've found that when idempotent PUT requests are retried, the bodies of the requests are empty because the `self.finished` attribute of the `S3Streamer` is evaluating to `true`. If the attribute to reset to `false` when `#rewind` is called, the body of the request is sent on retries.


geemus:
Thanks!

PR #209, opened 2015-12-19
Update dynamodb to use the latest API version

dmathieu:
The `20111205` API version is deprecated, and doesn't provide many of the features the latest version offers, such as the [conditional parameters](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LegacyConditionalParameters.html).

### DIFF
--- a/lib/fog/aws/requests/dynamodb/query.rb
+++ b/lib/fog/aws/requests/dynamodb/query.rb
@@ -6,15 +6,13 @@ module Fog
         #
         # ==== Parameters
         # * 'table_name'<~String> - name of table to query
-        # * 'hash_key'<~Hash> - hash key to query
         # * options<~Hash>:
         #   * 'AttributesToGet'<~Array> - Array of attributes to get for each item, defaults to all
         #   * 'ConsistentRead'<~Boolean> - Whether to wait for consistency, defaults to false
         #   * 'Count'<~Boolean> - If true, returns only a count of such items rather than items themselves, defaults to false
         #   * 'Limit'<~Integer> - limit of total items to return
-        #   * 'RangeKeyCondition'<~Hash>: value to compare against range key
-        #     * 'AttributeValueList'<~Hash>: one or more values to compare against
-        #     * 'ComparisonOperator'<~String>: comparison operator to use with attribute value list, in %w{BETWEEN BEGINS_WITH EQ LE LT GE GT}
+        #   * 'KeyConditionExpression'<~String> - the condition elements need to match
+        #   * 'ExpressionAttributeValues'<~Hash> - values to be used in the key condition expression
         #   * 'ScanIndexForward'<~Boolean>: Whether to scan from start or end of index, defaults to start
         #   * 'ExclusiveStartKey'<~Hash>: Key to start listing from, can be taken from LastEvaluatedKey in response
         #
@@ -25,15 +23,17 @@ module Fog
         #     * 'Count'<~Integer> - number of items in response
         #     * 'Items'<~Array> - array of items returned
         #     * 'LastEvaluatedKey'<~Hash> - last key scanned, can be passed to ExclusiveStartKey for pagination
-        def query(table_name, hash_key, options = {})
+        #
+        # See DynamoDB Documentation: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html
+        #
+        def query(table_name, options = {})
           body = {
-            'TableName'     => table_name,
-            'HashKeyValue'  => hash_key
+            'TableName'     => table_name
           }.merge(options)
 
           request(
             :body     => Fog::JSON.encode(body),
-            :headers  => {'x-amz-target' => 'DynamoDB_20111205.Query'}
+            :headers  => {'x-amz-target' => 'DynamoDB_20120810.Query'}
           )
         end
       end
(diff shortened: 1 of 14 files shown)


dmathieu:
Note that this includes several breaking changes. So we might want to release a major version with it, or change it to support both the deprecated and new versions.

lib/fog/aws/requests/dynamodb/query.rb
```diff
@@ -25,15 +23,17 @@ class Real
         #     * 'Count'<~Integer> - number of items in response
         #     * 'Items'<~Array> - array of items returned
         #     * 'LastEvaluatedKey'<~Hash> - last key scanned, can be passed to ExclusiveStartKey for pagination
-        def query(table_name, hash_key, options = {})
+        #
+        # See DynamoDB Documentation: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html
+        #
+        def query(table_name, options = {})
```
geemus: I'm not super familiar with the new stuff here, but is there any way we could take either format here? ie if hash_key is supplied, warn about deprecation but then merge it appropriately into the options and continue? If so, would be great to avoid the backwards-incompatibility here.

lib/fog/aws/requests/dynamodb/update_item.rb
```diff
@@ -6,39 +6,33 @@ class Real
...
         #
         # ==== Returns
         # * response<~Excon::Response>:
         #   * body<~Hash>:
         #     varies based on ReturnValues param, see: http://docs.amazonwebservices.com/amazondynamodb/latest/developerguide/API_UpdateItem.html
-        def update_item(table_name, key, attribute_updates, options = {})
+        #
+        # See DynamoDB Documentation: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateItem.html
+        #
+        def update_item(table_name, key, options = {})
```
geemus: Similar question here as to above. Backwards compatible would be preferable if we can do so reasonably.

lib/fog/aws/requests/dynamodb/query.rb
dmathieu: The previous API required a HashKeyValue argument, which contained: `{:S => 1}`.
See http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/API_Query_v20111205.html

DynamoDB would take the id and compare the provided value with it.
The new API requires to do:

```
:KeyConditionExpression => 'id = :id'
```

And unless the DynamoDB documentation is wrong, it doesn't appear `HashKeyValue` is still supported.
Transforming the former hash to the new value seems difficult, as we don't know the name of the primary key.

lib/fog/aws/requests/dynamodb/update_item.rb
dmathieu: DynamoDB appears to still support `AttributeUpdates`, though it's deprecated. So yes, we could be backward-compatible in the request.
I don't think the returned object would be though.

geemus: Returned object changes should probably be ok (aws tends to do this in an additive way). It is less of a break anyway.

lib/fog/aws/requests/dynamodb/query.rb
geemus: Hmm. Maybe we should still catch that case to provide a more explicit/specific error message, even if we just have to bail out at that point. What do you think?

dmathieu: Sure, that works for me. I'll do that.

dmathieu: I've updated this to show a deprecation notice and fallback to the former api version.

lib/fog/aws/requests/dynamodb/update_item.rb
dmathieu: Same here, updated to show a deprecation warning, but still support the older arguments.

geemus:
Thanks, looking good overall.

lib/fog/aws/requests/dynamodb/query.rb
```diff
@@ -25,7 +23,17 @@ class Real
         #     * 'Count'<~Integer> - number of items in response
         #     * 'Items'<~Array> - array of items returned
         #     * 'LastEvaluatedKey'<~Hash> - last key scanned, can be passed to ExclusiveStartKey for pagination
-        def query(table_name, hash_key, options = {})
+        #
+        # See DynamoDB Documentation: http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html
+        #
+        def query(table_name, options = {}, hash_key_deprecated = nil)
+          if has_deprecated_hash_key || (options.keys.length == 1 && [:S, :N, :B].include?(options.keys.first.to_sym))
```
geemus: Hey, maybe I'm just missing something, but where is `has_deprecated_hash_key` defined? Seems like maybe we could just drop that part of the conditional?

dmathieu: It is a method argument. We need it because the method signature changes, as DynamoDB doesn't have any `HashKey` option anymore. We need to do a string `KeyConditionExpression` instead.

geemus: Right. But in that case wouldn't it be `if hash_key_deprecated`?

dmathieu: You're absolutely right. I've fixed it. Thanks you.

geemus:
Thanks!

PR #210, opened 2015-12-21
Support NoncurrentVersion[Expiration,Transition] for s3 lifecycle.

xtoddx:
S3 has lifecycle policies for versioned objects that fog didn't support. This supports writing & parsing lifecycle specs for non-current versions of objects.

An example:
```ruby
require 'fog/aws'
s3 = Fog::Storage[:aws]
b = s3.directories.get('cmtestlifecycle')
lifecycle = { 'Rules' => [ { 'ID' => 'one-month-retention', 'Enabled' => true, 'NoncurrentVersionExpiration' => {'NoncurrentDays' => 30} } ] }
s3.put_bucket_lifecycle b.key, lifecycle
res = s3.get_bucket_lifecycle(b.key)
```


geemus:
Thanks!

Issue #208, opened 2015-12-10
Fog gives wrong location for buckets when connected via non-default region


jamesremuscat:
In this example, the bucket "jrem-test" has been created in the us-east-1 region.

```
>> Fog::Storage.new({provider: "AWS", region: "us-east-1"}).directories.get("jrem-test").location
"us-east-1"
```
as you'd expect. However, connecting via another region gives a different answer:

```
Fog::Storage.new({provider: "AWS", region: "eu-west-1"}).directories.get("jrem-test").location
[fog][WARNING] fog: followed redirect to jrem-test.s3-eu-west-1.amazonaws.com, connecting to the matching region will be more performant
"eu-west-1"
```

(Some operations on the bucket using the latter connection work fine, but (for example) directory.files.all throws an error complaining that the wrong region was used; creating some new objects fails with an error that suggests Fog isn't following redirects correctly - which is probably another issue).

I think the problem lies in lib/fog/aws/models/storage/directory.rb:34, where if no location constraint is set on an object, the region of the connection is used. I think this should instead use the AWS default region.

geemus:
Redirect following is definitely not working right, we probably need to change it to do more requests to figure things out properly. Maybe falling back to default is the better choice here, would you be up for working on a related pull request?

geemus:
Fixed by #211

PR #211, opened 2015-12-31
When not specified, region for a bucket should be DEFAULT_REGION.

jamesremuscat:
As detailed in #208, buckets without location constraints should default to the AWS default region and not that used by the current connection.


jamesremuscat:
Some of the automated tests are failing, but don't seem to be connected to my (very limited) change.

geemus:
Thanks!

PR #215, opened 2016-01-14
add mime-type.gem

marcy:
The mime-types gem is needed. https://github.com/fog/fog-aws/commit/13bda13933711c7d3efa00bcf1e913e053aa34e6

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,5 @@ gemspec
 
 group :test, :default do
   gem 'pry-nav'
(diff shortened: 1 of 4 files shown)


geemus:
I'm reticent to add it as a general dependency because it has locked versions for earlier rubies (as you ran into). Also it is not even needed for users of parts other than storage. As such we are (at least for now) trying to simply use the requires you mention as a way to make it an optional requirement where end users can manage the version appropriately when they need it. Does that help/make sense? Thanks!

PR #217, opened 2016-01-19
restore db instance fom db snapshot

ehowe:
* add mock
* add restore method on servers collection
* modify save method on server model
* raise error when multi_az and az are set
* cannot specify multi_az to be true and pick an az together
* take allocated_storage from snapshot on restore
* restore from snapshot endpoint does not take
* allocated_storage parameter
* save storage type on rds snapshot

### DIFF
--- a/fog-aws.gemspec
+++ b/fog-aws.gemspec
@@ -28,4 +28,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'fog-json',  '~> 1.0'
   spec.add_dependency 'fog-xml',   '~> 0.1'
   spec.add_dependency 'ipaddress', '~> 0.8'
+  spec.add_dependency 'mime-types'
 end
--- a/fog-aws.gemspec
+++ b/fog-aws.gemspec
@@ -28,5 +28,4 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'fog-json',  '~> 1.0'
   spec.add_dependency 'fog-xml',   '~> 0.1'
   spec.add_dependency 'ipaddress', '~> 0.8'
-  spec.add_dependency 'mime-types'
 end
--- a/lib/fog/aws/dns.rb
+++ b/lib/fog/aws/dns.rb
@@ -3,18 +3,6 @@ module Fog
     class AWS < Fog::Service
       extend Fog::AWS::CredentialFetcher::ServiceMethods
 
-      require 'fog/aws/requests/dns/change_resource_record_sets'
-      require 'fog/aws/requests/dns/create_health_check'
-      require 'fog/aws/requests/dns/create_hosted_zone'
-      require 'fog/aws/requests/dns/delete_health_check'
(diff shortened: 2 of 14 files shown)


ehowe:
Not sure whats up with the specs, every time I fix something, something else breaks.  Is this a known issue?

lib/fog/aws/dns.rb
```diff
@@ -3,6 +3,18 @@ module DNS
     class AWS < Fog::Service
       extend Fog::AWS::CredentialFetcher::ServiceMethods
 
+      require 'fog/aws/requests/dns/change_resource_record_sets'
```
geemus: These should be redundant to the `request` calls that come a bit later (though a couple new ones should be added). By using `request` the requires are deferred, however, until initialization time. So, I think we just want the request versions and not these explicit requires.

fog-aws.gemspec
```diff
@@ -28,4 +28,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'fog-json',  '~> 1.0'
   spec.add_dependency 'fog-xml',   '~> 0.1'
   spec.add_dependency 'ipaddress', '~> 0.8'
+  spec.add_dependency 'mime-types'
```
geemus: My preference is to keep this out of the fog/fog-aws gemspec (since different versions are needed depending on which ruby you are using). We can have a guarded/warning if you try to do something that needs it without having it loaded yet.

lib/fog/aws/models/rds/server.rb
```diff
@@ -104,16 +104,22 @@ def modify(immediately, options)
...
-          requires :allocated_storage
-          requires :master_username
-          requires :password
-
-          self.flavor_id ||= 'db.m1.small'
-
-          data = service.create_db_instance(id, attributes_to_params)
-          merge_attributes(data.body['CreateDBInstanceResult']['DBInstance'])
-          true
+          if source_snapshot_id
```
geemus: Since this is pretty different, maybe we should just have a separate restore function (instead of having save pull double duty). I suspect that might make it clearer and less error prone for users. What do you think?

geemus:
Looks good overall, added a couple more specific items of feedback above. Thanks!

lib/fog/aws/models/rds/server.rb
jacobo: Calling it restore is a probably a bit of a gaff on amazon's part, as it creates an entirely new server. There is no true "restore existing" functionality, you are just expected to create replacements.
 
The behavior of this `save` method therefore is consistent with `Volume#save` / `Volume.create` (optionally taking a snapshot_id).

see: http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBInstanceFromDBSnapshot.html

geemus: Ah, yeah, that makes sense. Thanks for clarifying.

geemus:
Thanks!

PR #221, opened 2016-01-28
Fix for bug where retrying a chunked put_object request would fail.

Sumo-MBryant:
If a chunked put_object request fails, all subsequent retries also fail due to the "finished" flag not being reset. Subsequent attempts upload 0 bytes which AWS rejects as not matching the header.


geemus:
Anything I can help with or clarify here?

PR #229, opened 2016-02-24
autoscaler attach/detatch

shaiguitar:
Ran this against a live amazon account and this did the expected. 

Amazon doc links inline in the commit.

### DIFF
--- a/lib/fog/aws/requests/auto_scaling/attach_instances.rb
+++ b/lib/fog/aws/requests/auto_scaling/attach_instances.rb
@@ -16,7 +16,7 @@ module Fog
         #
         # ==== See Also
         #
-        # http://docs.aws.amazon.com/cli/latest/reference/autoscaling/attach-instances.html
+        # http://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_AttachInstances.html
 
         ExpectedOptions[:asg_name] = %w[AutoScalingGroupName]
         ExpectedOptions[:instance_ids] = %w[InstanceIds]
--- a/lib/fog/aws/requests/auto_scaling/attach_instances.rb
+++ b/lib/fog/aws/requests/auto_scaling/attach_instances.rb
@@ -1,0 +1,59 @@
+module Fog
+  module AWS
+    class AutoScaling
+      class Real
+        require 'fog/aws/parsers/auto_scaling/basic'
+
+        # Removes one or more instances from the specified Auto Scaling group.
+        #
+        #  cli equiv:
+        # `aws autoscaling attach-instances --instance-ids i-93633f9b --auto-scaling-group-name my-auto-scaling-group`
+        #
+        # ==== Parameters
+        #
+        #   * AutoScalingGroupName<~String> - The name of the Auto Scaling group``
+        #   * 'InstanceIds'<~Array> - The list of Auto Scaling instances to detach.
+        #
+        # ==== See Also
+        #
+        # http://docs.aws.amazon.com/cli/latest/reference/autoscaling/attach-instances.html
+
+        ExpectedOptions[:asg_name] = %w[AutoScalingGroupName]
+
+        def attach_instances(auto_scaling_group_name, options = {})
+
+          if instance_ids = options.delete('InstanceIds')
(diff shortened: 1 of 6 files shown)


lib/fog/aws/requests/auto_scaling/attach_instances.rb
```diff
@@ -0,0 +1,60 @@
...
+        # `aws autoscaling attach-instances --instance-ids i-93633f9b --auto-scaling-group-name my-auto-scaling-group`
+        #
+        # ==== Parameters
+        #
+        #   * AutoScalingGroupName<~String> - The name of the Auto Scaling group``
+        #   * 'InstanceIds'<~Array> - The list of Auto Scaling instances to detach.
+        #
+        # ==== See Also
+        #
+        # http://docs.aws.amazon.com/cli/latest/reference/autoscaling/attach-instances.html
```
lanej: should be http://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_AttachInstances.html

lib/fog/aws/requests/auto_scaling/detach_instances.rb
```diff
@@ -0,0 +1,69 @@
...
+        # ==== Parameters
+        #
+        #   * AutoScalingGroupName<~String> - The name of the Auto Scaling group``
+        #   * 'InstanceIds'<~Array> - The list of Auto Scaling instances to detach.
+        #
+        # * options<~Hash>:
+        # 'shouldDecrementDesiredCapacity'<~Boolean> - decrement the asg capacity or not (it will boot another if an instance id detached)
+        #
+        # ==== See Also
+        # http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/detach-instance-asg.html
```
lanej: http://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_DetachInstances.html

shaiguitar:
pushed doc link change

lib/fog/aws/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fog
   module AWS
-    VERSION = "0.8.1"
+    VERSION = "0.9.0"
```
lanej: i'll worry about the version bump, tagging, etc.

shaiguitar:
reverted lib/version.rb change to what it was

geemus:
@lanej looks like you are on top of this, but just let me know if you need anything.

lanej:
@geemus thanks

lanej:
@shaiguitar the mocks don't manipulate any data.  is that intentional ?

shaiguitar:
@lanej Basically just copied over what was in https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/auto_scaling/attach_load_balancers.rb#L36-L55 . There's no documentation/wiki on adding mock behavior and it's been a while so I'm a little hazy on the idiomatics.

Seems like the asg mock interactions are broken as well with regards to the other instance manipulation requests, see these two (`Fog::Mock.not_implemented`) `lib/fog/aws/requests/auto_scaling/set_instance_health.rb` and `lib/fog/aws/requests/auto_scaling/terminate_instance_in_auto_scaling_group.rb`

In the case of `lib/fog/aws/requests/auto_scaling/set_desired_capacity.rb` if we reload after setting desired capacity it should also create some instances, but nothing is happening in the mock either.

There's probably some other examples but that should be an indication it's not quite there for the asgs anyway.

So, I can add another `not_implemented` in there or I can take a look at implementing these mocks, but I'll need some hand holding and or docs/wiki.

shaiguitar:
Bump. The real implementation should work. I can add in `not_implemented` if that's better than the current alternative

shaiguitar:
Anything I can do to get this released?

lanej:
@shaiguitar rebase for a passing build

shaiguitar:
@lanej green

shaiguitar:
Thanks!

lanej:
this commit screwed with the versions @2456533f029698f060f4c368af22d9106d74b64f.  Next time remove the commit instead of reverting in the same branch.

PR #233, opened 2016-03-02
Update README.md

h0lyalg0rithm:
Use high res badge


lanej:
Thanks.  Never noticed how bad the old one was.

geemus:
Thanks!

PR #241, opened 2016-04-13
Add support for CloudFormation stack tags.

### LINKED ISSUE #240: CloudFormation stack tags cause describe_stacks to break
When using CloudFormation stacks, the [`DescribeStacks` API action](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStacks.html) returns a list of stacks and information about them.

This information also includes [tags](http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Tag.html). However, fog does not return these tags; furthermore, _the presence of tags in the AWS API response_ breaks the response to `describe_stacks`.

This is because the `<Tags>` element of the response contains `<member>` elements. The `Fog::Parsers::AWS::CloudFormation::DescribeStacks` parser assumes that any `</member>` outside of `<Outputs>`, `<Parameters>` and `<Capabilities>` is the end of a stack description, so starts a new stack.

The net result is that a single stack in the AWS response gets turned into multiple 'stacks' in the Fog parsed result, many of which will be empty, and the `Outputs` and `Parameters` will be split across different 'stacks' (since `Tags` appears between them in the XML).

I suspect that adding support for returning tag information will solve this problem.

jamesremuscat:
The presence of tags on CloudFormation stacks would break the previous implementation due to the unexpected `<member>` elements.

Fixes #240.

### DIFF
--- a/lib/fog/aws/parsers/cloud_formation/describe_stacks.rb
+++ b/lib/fog/aws/parsers/cloud_formation/describe_stacks.rb
@@ -4,9 +4,10 @@ module Fog
       module CloudFormation
         class DescribeStacks < Fog::Parsers::Base
           def reset
-            @stack = { 'Outputs' => [], 'Parameters' => [], 'Capabilities' => [] }
+            @stack = { 'Outputs' => [], 'Parameters' => [], 'Capabilities' => [], 'Tags' => [] }
             @output = {}
             @parameter = {}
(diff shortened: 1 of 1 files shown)


geemus:
Oops, didn't notice this when I asked about you helping on #240, thanks for your help!

PR #245, opened 2016-04-20
Parse CloudWatch alarm actions as arrays instead of strings

eherot:
If someone else is willing to do the work to create a CloudWatch unit test framework I will write a test for this change. ;-)


geemus:
Thanks!

PR #252, opened 2016-05-11
do not make requests if mocked.

shaiguitar:
Doing a `compute.servers` should work according as expected in mock mode when using it with an iam role (ie `use_iam_profile`), but currently the credentials_fetcher is making real requests. This just mocks the api keys that it would get in the real mode.

### DIFF
--- a/tests/credentials_tests.rb
+++ b/tests/credentials_tests.rb
@@ -1,5 +1,6 @@
 Shindo.tests('AWS | credentials', ['aws']) do
   old_mock_value = Excon.defaults[:mock]
+  fog_was_mocked = Fog.mocking?
   Excon.stubs.clear
   Fog.mock!
   begin
@@ -52,5 +53,6 @@ Shindo.tests('AWS | credentials', ['aws']) do
   ensure
     Excon.stubs.clear
     Excon.defaults[:mock] = old_mock_value
+    Fog.unmock! if !fog_was_mocked
   end
 end
--- a/tests/credentials_tests.rb
+++ b/tests/credentials_tests.rb
@@ -1,7 +1,7 @@
 Shindo.tests('AWS | credentials', ['aws']) do
   old_mock_value = Excon.defaults[:mock]
   Excon.stubs.clear
-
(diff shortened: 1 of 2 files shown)


geemus:
It looks like there must be some expectation around this being mocked already? The tests appear to be looking for some specific dummy keys/secrets that are now failing. Could you take another look? Thanks!

shaiguitar:
Looks like the request is just being stubbed in the test, it's not happening Fog.mock level, which also verifies my experience of seeing it make requests even in mocked mode. Excon request is being patched up here: https://github.com/fog/fog-aws/blob/master/tests/credentials_tests.rb#L18
with this info https://github.com/fog/fog-aws/blob/master/tests/credentials_tests.rb#L11-L16

Can we just remove the Excon stub and default to using Fog.mock in the test itself? If so I'll commit a fix.

geemus:
Ah, thanks. I saw the failure but didn't dig much deeper. Removing that stub and using the mock stuff itself sounds good. Thanks!

shaiguitar:
Fix pushed 📗 The stubs are actually a bit more comprehensive tests so keeping them around for debugging if necessary later.

tests/credentials_tests.rb
```diff
@@ -1,7 +1,7 @@
 Shindo.tests('AWS | credentials', ['aws']) do
   old_mock_value = Excon.defaults[:mock]
   Excon.stubs.clear
-
+  Fog.mock!
```
lanej: this will pollute the rest of the test suite if not running mocked already

shaiguitar: 84315d9d

lanej:
💥

geemus:
Thanks!

jeffjo:
can you cut a new release with this change?

jeffjo:
Woops, I commented in the wrong issue. I was referring to the fix for #374.

geemus:
@jeffjo ah ha, I suspected it may have just been a mistake on the target. I'll comment over on that issue.

PR #257, opened 2016-06-12
Fix for empty ETag values

baryshev:
Fog aws fails if API response contains empty ETag tags. It may cause when working with S3 clones (like minio in my case), that not provides ETags in their API answer.


lanej:
Thanks @baryshev

geemus:
Thanks!

PR #258, opened 2016-06-13
Allow case-insensitive record comparison

mpick92:
I want to change to case-insensitive comparison to allow this use case:
```
"server1.example.com".casecmp("SERVER1.example.com") == 0
```


lanej:
https://tools.ietf.org/html/rfc4343#section-2

lanej:
Thanks @mpick92

geemus:
Thanks!

PR #259, opened 2016-06-17
Parse EbsOptimized parameter in launch configuration description

djudd:
This value is always treated as nil in current master

As in a similar prior pull, I'm not sure where/if to add tests: https://github.com/fog/fog-aws/pull/29


geemus:
Thanks!

PR #262, opened 2016-06-21
[Do Not Merge] AWS CloudFront (CDN) API Upgrade

### LINKED ISSUE #17: Bring AWS CloudFront API Models/Requests up to date
Original issue at: fog/fog#3115

Heaps of missing functionality, specifically around the use of Price Classes, Custom/Dedicated SSL certificates, etc.

Current API version is 3~ years behind, has some authentication breakage when you try and just change the version to 2014-05-31:

```
 response => #<Excon::Response:0x007f991daab9b0 @data={:body=>"<?xml version=\"1.0\"?>\n<ErrorResponse xmlns=\"http://cloudfront.amazonaws.com/doc/2014-01-31/\"><Error><Type>Sender</Type><Code>IncompleteSignature</Code><Message>Authorization header requires 'Credential' parameter. Authorization header requires 'Signature' parameter. Authorization header requires 'SignedHeaders' parameter.
```

johnnyshields:
Fixes #17

I've upgraded the authentication scheme for CloudFront and it seems to be working. However, I'm currently getting stuck with the following error:

```
Excon::Errors::SocketError: Opening and ending tag mismatch: OriginAccessIdentity line 0 and S3OriginConfig (Nokogiri::XML::SyntaxError)
```

(Possibly AWS is sending malformed XML in the response?)

In addition, there will be changes to the object model required. FYI:
- New Docs: http://docs.aws.amazon.com/AmazonCloudFront/latest/APIReference/
- Old Docs: http://docs.aws.amazon.com/AmazonCloudFront/2010-11-01/APIReference/

### DIFF
--- a/lib/fog/aws/cdn.rb
+++ b/lib/fog/aws/cdn.rb
@@ -55,10 +55,6 @@ module Fog
           self.class.data.delete(@aws_access_key_id)
         end
 
-        def signature(params)
-          "foo"
-        end
-
         def setup_credentials(options={})
           @aws_access_key_id  = options[:aws_access_key_id]
         end
@@ -106,7 +102,7 @@ module Fog
           response = Excon::Response.new
           response.status = status
           response.body = <<EOF
-<ErrorResponse xmlns="http://cloudfront.amazonaws.com/doc/2010-11-01/">
+<ErrorResponse xmlns="http://cloudfront.amazonaws.com/doc/2016-01-13/">
    <Error>
       <Type>Sender</Type>
       <Code>#{code}</Code>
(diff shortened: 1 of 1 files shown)


geemus:
I believe if you tweak the request by commenting out the parser and then run the command, you should get back the raw response. That should hopefully make it a bit more clear what you are seeing and then we can work back from there.

fcheung:
You can also set the DEBUG_RESPONSE environment variable which should log it all to wherever Fog::Logger is pointed

geemus:
@fcheung yeah, I think it might be weird in this case (as the response is streamed to the parser, I don't think it shows up in full on the debug output). Or at least that is what I thought I remembered...

johnnyshields:
At the moment I don't have the bandwidth to work on this further. I do think I've made significant progress with the authorization bit, but there's more to do including updates to the object models. If someone who is familiar with fog can take this over it would be much appreciated.

PR #263, opened 2016-06-21
AWS DNS - support newer DNS hosted zone IDs for dualstack ELBs

mattheworiordan:
See https://forums.aws.amazon.com/message.jspa?messageID=612414

Setting up DNS ALIAS records for the newer (and now default) dualstack DNS records without the correct hosted zone ID will result in odd things happening, typically only one IP address being returned for an ELB.


geemus:
Thanks!

mattheworiordan:
No prob :)

Issue #261, opened 2016-06-20
Digest::Base Error


youreddy:
Hello, 

I'm getting a `Digest::Base cannot be directly inherited in Ruby` when using this the fog-aws storage. Here's the stack trace from the error: 

```
  "Request failed: 500: {\"code\"=>10001, \"description\"=>\"Digest::Base cannot be directly inherited in Ruby\", \"backtrace\"=>[
  \"vendor/bundle/ruby/2.3.0/gems/fog-aws-0.9.1/lib/fog/aws/storage.rb:588:in `digest'\", 
  \"vendor/bundle/ruby/2.3.0/gems/fog-aws-0.9.1/lib/fog/aws/storage.rb:588:in `hexdigest'\", 
  \"vendor/bundle/ruby/2.3.0/gems/fog-aws-0.9.1/lib/fog/aws/storage.rb:588:in `request'\",
   \"vendor/bundle/ruby/2.3.0/gems/fog-aws-0.9.1/lib/fog/aws/requests/storage/get_bucket.rb:43:in `get_bucket'\",
   \"vendor/bundle/ruby/2.3.0/gems/fog-aws-0.9.1/lib/fog/aws/models/storage/directories.rb:21:in `get'\",
```

From a quick google search, it looks like this is a reoccurring issue in Digest that others have solved used `OpenSSL::Digest` instead. 

Here are some links to related issues that might be helpful: 
https://github.com/aws/aws-sdk-ruby/pull/663
https://github.com/chef/chef/pull/1807

geemus:
Thanks for the heads up, would you be up for helping on a pull request? Thanks!

geemus:
Closed by #266

PR #266, opened 2016-06-28
Replaces usage of Digest with OpenSSL::Digest 

esthervillars:
The Digest class has some thread safety issues and gives runtime error 
Digest::Base cannot be directly inherited in Ruby #261

ruby/ruby@c02fa39
aws/aws-sdk-ruby#525
aws/aws-sdk-ruby#529


lanej:
@geemus this looks fine to me.

geemus:
Looks good, thanks!

PR #269, opened 2016-06-30
Rds snapshot improvements

tekken:
Implemented API for ModifyDBSnapshotAttribute and CopyDBSnapshot


lib/fog/aws/requests/rds/copy_db_snapshot.rb
```diff
@@ -0,0 +1,53 @@
+module Fog
+  module AWS
+    class RDS
+      class Real
+        require 'fog/aws/parsers/rds/create_db_snapshot'
```
geemus: Is this supposed to re-use create parser, or should it have it's own?

geemus:
Looks like a great start, one comment inline and would be great to also see some tests more fleshed out. Thanks!

tests/requests/rds/db_snapshot_tests.rb
```diff
@@ -0,0 +1,73 @@
...
+  @snapshots_format = {
+    'requestId'   => String
+  }
+
+  @snapshot_copy_result = {
+    'requestId'   => String,
+    'snapshotId'  => String
+  }
+
+  Fog.mock!
```
tekken: I am unsure if this is required though i couldnt pass the test cases without it..throws a
` Missing required arguments: aws_access_key_id, aws_secret_access_key (ArgumentError)`

lib/fog/aws/requests/rds/copy_db_snapshot.rb
tekken: Your right it should have its own.

tests/requests/rds/db_snapshot_tests.rb
geemus: I don't think this should be here (it would be set or not at a higher level). The test you mention is broken on master I'm pretty sure, so we can consider this good even without this. If you can remove this I think we should be good to merge the rest. Thanks!

tekken: Ok. I have removed it.

geemus:
Looks good, thanks!

Issue #272, opened 2016-07-06
Feature: Custom Managed Policies


MrPrimate:
I would like support to update Custom Managed IAM Policies.

`create_policy` can create policies, however the policy model expects a user or group policy, and no support for updating policy/updating policy version with CreatePolicyVersion API call.

geemus:
Sounds good. Could you sketch out what you imagine the end result would look like and/or how it would work? Hopefully we can reach some consensus there before getting too far into code and it can guide the implementation. Thanks!

geemus:
oops, accidental close, just meant to comment

MrPrimate:
I think IAM requests to CreatePolicyVersion to create a new policy version, SetDefaultPolicyVersion to set a default policy, and DeletePolicyVersion to delete a policy version would allow full manipulation of Custom Managed IAM Policies.

I'm not quite sure how to work it into the model though.

geemus:
I'm afraid I don't have a lot of familiarity with this space, so I'm happy to listen and discuss but not sure I have much to offer without help.

MrPrimate:
@geemus, I think something like https://github.com/fog/fog-aws/pull/274, although I think I will need assistance with the tests. Thanks.

geemus:
Awesome. I'm going to go ahead and close this in favor of consolidating discussion on #274.

PR #274, opened 2016-07-11
Expanding IAM support

MrPrimate:
Adding:
* Create/Delete Policy Versions
* Set Default Policy Version
* Add Update Assume Role Policy

For functionality request https://github.com/fog/fog-aws/issues/272

### DIFF
--- a/lib/fog/aws/iam.rb
+++ b/lib/fog/aws/iam.rb
@@ -61,6 +61,7 @@ module Fog
       request :list_access_keys
       request :list_account_aliases
       request :list_attached_group_policies
+      request :list_attached_role_policies
       request :list_attached_user_policies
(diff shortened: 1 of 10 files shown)


geemus:
Looking good so far. As for tests, I think a good starting point would be just some CRUD type stuff that creates/updates/sets/deletes policy would be a great place to start. Ideally it would pass vs all the mocks (and you could run it live to ensure that this also works and that mock/real behavior align). Does that make sense? Let me know if I can help/clarify further. Thanks!

MrPrimate:
First go at tests, and is working as desired on my test AWS account to create, update and delete policy versions.

geemus:
Looks good, thanks!

PR #276, opened 2016-07-13
RDS test fixes

MrPrimate:
Some RDS tests were failing due to a bad mock returned with copy_db_snapshot, and a couple of old/incorrect formats for said snapshots.


geemus:
Awesome, I noticed this yesterday and was intending to dig in to it, but hadn't quite gotten that far. Thanks!

PR #290, opened 2016-08-17
Add gestion of egress security group rules

KevinLoiseau:
Hi guys !

Just a little PR to add gestion of egress security group rules (authorize / revoke).

For ensure retro-compatibility I add an optional parameters that indicate if the rules is ingress or egress, but then it's less better abstraction of AWS API. WDYT about this ?

I'm not sure that is the best way for implement this functionality, I will be listening you, feel free to tell me if I'm doing something wrong.
Thanks.

### DIFF
--- a/tests/models/compute/security_group_tests.rb
+++ b/tests/models/compute/security_group_tests.rb
@@ -38,13 +38,25 @@ Shindo.tests("Fog::Compute[:aws] | security_group", ['aws']) do
     test("authorize access at a port range (egress rule)") do
       @group.authorize_port_range(5000..6000, :direction => 'egress')
       @group.reload
-      @group.ip_permissions_egress.size == 1
+      ip_permission_egress = @group.ip_permissions_egress.find do |permission|
+        permission['fromPort'] == 5000 &&
+          permission['toPort'] == 6000 &&
+          permission['ipProtocol'] == 'tcp' &&
+          permission['ipRanges'] == [{ 'cidrIp' => '0.0.0.0/0' }]
+      end
+      !ip_permission_egress.nil?
     end
 
     test("revoke access at a port range (egress rule)") do
       @group.revoke_port_range(5000..6000, :direction => 'egress')
       @group.reload
-      @group.ip_permissions_egress.empty?
+      ip_permission_egress = @group.ip_permissions_egress.find do |permission|
+        permission['fromPort'] == 5000 &&
+          permission['toPort'] == 6000 &&
+          permission['ipProtocol'] == 'tcp' &&
+          permission['ipRanges'] == [{ 'cidrIp' => '0.0.0.0/0' }]
+      end
+      ip_permission_egress.nil?
     end
 
     group_forms = [
--- a/tests/models/compute/security_group_tests.rb
+++ b/tests/models/compute/security_group_tests.rb
@@ -35,6 +35,12 @@ Shindo.tests("Fog::Compute[:aws] | security_group", ['aws']) do
       @group.ip_permissions.empty?
     end
 
+    test("authorize access at a port range (egress rule)") do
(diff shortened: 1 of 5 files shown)


geemus:
Looks like a reasonable approach to me. Thanks!

@lanej could you also take a peek and merge if it seems reasonable to you? Thanks!

tests/models/compute/security_group_tests.rb
```diff
@@ -35,6 +35,18 @@
       @group.ip_permissions.empty?
     end
 
+    test("authorize access at a port range (egress rule)") do
+      @group.authorize_port_range(5000..6000, :direction => 'egress')
+      @group.reload
+      @group.ip_permissions_egress.size == 1
```
lanej: in reality, there are two rules here

```
[2] pry(#<Shindo::Tests>)> @group.ip_permissions_egress
=> [{"groups"=>[], "ipRanges"=>[{"cidrIp"=>"0.0.0.0/0"}], "ipProtocol"=>"tcp", "fromPort"=>5000, "toPort"=>6000}, {"groups"=>[], "ipRanges"=>[{"cidrIp"=>"0.0.0.0/0"}], "ipProtocol"=>"-1"}]
```

I would recommend changing the assertion to verify that the rule you created is present, rather then asserting absolute size.

tests/models/compute/security_group_tests.rb
```diff
@@ -35,6 +35,18 @@
...
+    end
+
+    test("revoke access at a port range (egress rule)") do
+      @group.revoke_port_range(5000..6000, :direction => 'egress')
+      @group.reload
+      @group.ip_permissions_egress.empty?
```
lanej: similar to https://github.com/fog/fog-aws/pull/290/files#r75165909, i would recommend asserting the absence of the specified rule.

KevinLoiseau:
@geemus, @lanej Thanks for your feedback !

@lanej I tried to apply your recommendation, but with my implementation if we change the default value of 'ipPermission' or 'ipRange' fields this test will fail then he should not. WDYT about this ?

Thanks.

tests/models/compute/security_group_tests.rb
```diff
@@ -35,6 +35,32 @@
       @group.ip_permissions.empty?
     end
 
+    test("authorize access at a port range (egress rule)") do
+      @group.authorize_port_range(5000..6000, :direction => 'egress')
+      @group.reload
+      ip_permissions_egress = @group.ip_permissions_egress.dup
+      ip_permissions_egress.keep_if do |permission|
```
lanej: it would be easier (and 1.8.7 compatible) to use `@group.ip_permissions_egress.find { }` and allow that to be the assertion.

tests/models/compute/security_group_tests.rb
```diff
@@ -35,6 +35,32 @@
...
+          permission['ipRanges'] == [{ 'cidrIp' => '0.0.0.0/0' }]
+      end
+      ip_permissions_egress.size == 1
+    end
+
+    test("revoke access at a port range (egress rule)") do
+      @group.revoke_port_range(5000..6000, :direction => 'egress')
+      @group.reload
+      ip_permissions_egress = @group.ip_permissions_egress.dup
+      ip_permissions_egress.keep_if do |permission|
```
lanej: same here https://github.com/fog/fog-aws/pull/290/files#r75333885

nictrix:
Hi, I was looking for managing egress authorization/revoking rules for security groups in AWS, this PR implements that ability? If so is there any plans to merge it into the next release of the gem? Looking forward to having this ability!

KevinLoiseau:
@nictrix Yes, this PR implement this abilities.

@lanej I had update my test to use `find` instead of `keep_if`, that is sound good to you ?

lanej:
@KevinLoiseau `find` works for me.

PR #297, opened 2016-09-13
Cloud Formation: support for change sets, stack policy and other missing calls.  

neillturner:
This adds all the missing commands from the latest aws cloud formation including support for change sets, stack policy and other missing cloud formation calls.  This is mostly additional code it would be great if we can get this in next version.


geemus:
Thanks!

PR #298, opened 2016-09-14
Cloud Formation: additional parameters

neillturner:
Some additional parameters I missed for CreateStack and UpdateStack


geemus:
Thanks!

PR #299, opened 2016-09-20
Add attribute is_default in vpc

zhitongLBN:
Add `is_default` attribute in vpc model

We have this attribute in aws api response and we saved this attribute in response parser as a string but not with a real attribute


geemus:
Thanks!

PR #302, opened 2016-09-26
Correct optional parameter naming in documentation for Fog::AWS::Auto…

ehealy:
…scaling::Real#put_scaling_policy

Stumbled onto this when working with the put_scaling_policy method. Just correcting the documentation to match http://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_PutScalingPolicy.html


geemus:
Thanks!

PR #306, opened 2016-10-07
Put accelerated object url

### LINKED ISSUE #303: Fog transfer acceleration endpoints
When using `Fog::Storage` with `AWS`

Method `put_object_url` does not support [AWS Transfer Acceleration](https://aws.amazon.com/blogs/aws/aws-storage-update-amazon-s3-transfer-acceleration-larger-snowballs-in-more-regions/).

Usually, you can you can upload files to `https://s3.amazonaws.com/bucketname`, however, with transfer acceleration, it needs to be `https://bucketname.s3-accelerate.amazonaws.com`, hence the parameter [`bucket_name`](https://github.com/fog/fog/blob/master/lib/fog/internet_archive/requests/storage/put_object_url.rb#L18) becomes superfluous

ilyakatz:
Fixes #303

### DIFF
--- a/lib/fog/aws/requests/storage/put_accelerated_object_url.rb
+++ b/lib/fog/aws/requests/storage/put_accelerated_object_url.rb
@@ -1,0 +1,39 @@
+module Fog
+  module Storage
+    class AWS
+      module PutAcceleratedObjectUrl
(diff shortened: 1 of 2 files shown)


geemus:
Could you add a test or two to make sure this does what we want (and doesn't break in the future)? Thanks!

PR #310, opened 2016-10-15
fix S3 #delete_multiple_objects for UTF-8 names

alepore:
Hello!
I was debugging https://github.com/backup/backup/issues/670 and i found out that fog-aws' `#delete_multiple_objects` does not work with UTF-8 file names.

The problem is in the 'Content-Length' header, `String#length` is not the right method to find UTF-8 strings size:

```
> "fØg_öbjèct".length
 => 10
> "fØg_öbjèct".bytesize
 => 13
```

Please review the test, it not very clear to me if that's enough (https://github.com/fog/fog-aws/issues/311)

### DIFF
--- a/lib/fog/aws/requests/storage/delete_multiple_objects.rb
+++ b/lib/fog/aws/requests/storage/delete_multiple_objects.rb
@@ -46,7 +46,7 @@ module Fog
           end
           data << "</Delete>"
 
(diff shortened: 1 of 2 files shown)


lanej:
@alepore good catch.

geemus:
Thanks!

PR #312, opened 2016-10-16
mime types gem update

lucianosousa:


geemus:
I think this might break our compatibility with ruby < 2.0? We need to do that at some point anyway, but I think we are still holding out so far. I am uncertain though, there were some related issues previously (such as https://github.com/fog/fog-core/pull/170). @lanej maybe you can help me recall and figure out the impact of this?

lucianosousa:
@geemus there is any test case related to this? seems everything is fine because tests are passing

lanej:
@geemus because we don't put `mime-types` in the Gemfile, the only issue here is that we are missing test coverage for mime-types 2.x.  It maybe that we want to make some additional Gemfiles for for ruby > 2.0 and `mime-types` ~> 2.0 for CI to crunch on.

lucianosousa:
I can be wrong but looks like mime-types gem is required in all rubies on CI
https://travis-ci.org/fog/fog-aws/builds/168120986

geemus:
I think I was confusing myself anyway. The issue I was thinking about would come up if we changed the gemspec (but changing the Gemfile should be fine, and testing against latest seems reasonable). Going to bring this in, sorry for the confusion and thanks!

lucianosousa:
not a problem. tks for the thoughts

PR #314, opened 2016-10-18
Ohio region

chanakyacool:
+ Added region 'us-east-2' ohio
+ Added Az's 
 - us-east-2a
 - us-east-2b
 - us-east-2c


geemus:
Thanks!

PR #318, opened 2016-11-04
data pipeline mocks

ehowe:


geemus:
Broadly looks good to me (I have no experience with pipeline stuff though).

@lanej I'll leave it to you? Thanks.

lanej:
@ehowe ready ?

ehowe:
Yup, let's do it.

lib/fog/aws/requests/data_pipeline/put_pipeline_definition.rb
```diff
@@ -64,11 +36,60 @@ def field_for_kv(key, value)
...
+        # ==== Returns
+        # * response<~Excon::Response>:
+        #   * body<~Hash>:
+        def put_pipeline_definition(id, pipeline_objects, options={})
+          params = {
+            'pipelineId' => id,
+            'pipelineObjects' => transform_objects(pipeline_objects),
+          }.merge(options)
+
+          response = request({
```
lanej: nitpick: unnecessary assignment.  couple of these lying around.

lib/fog/aws/requests/data_pipeline/activate_pipeline.rb
```diff
@@ -14,16 +14,20 @@ def activate_pipeline(id)
 
           response = request({
             :body => Fog::JSON.encode(params),
-            :headers => { 'X-Amz-Target' => 'DataPipeline.ActivatePipeline' },
+            :headers => { 'X-Amz-Target' => 'DataPipeline.ActivatePipeline' }
           })
-
-          Fog::JSON.decode(response.body)
```
lanej: We are changing the return value for all of these requests to the raw response from the decoded body.  While I think this is good direction, I would like @fog/fog-aws to comment on the relevance of the breaking change.

ehowe: Definitely brings it in line with the rest of the services.  Without making that change, it makes it impossible to assert on things that return an empty body, but different status codes.

lanej:
@geemus this PR changes the return value of the data pipeline requests.  Would you consider that a breaking change constituting a major version bump ?  Do we consider request to be part of the public API ?

lib/fog/aws/data_pipeline.rb
```diff
@@ -116,7 +167,19 @@ def request(params)
         end
 
         def _request(params)
-          @connection.request(params)
+          response = @connection.request(params)
+
+          unless response.body.empty?
```
geemus: @lanej Won't this part mean that we still end up with decoded bodies? Or am I misunderstanding your question?

lib/fog/aws/requests/data_pipeline/create_pipeline.rb
```diff
@@ -24,14 +24,35 @@ def create_pipeline(unique_id, name, description=nil, tags=nil)
             :body => Fog::JSON.encode(params),
             :headers => { 'X-Amz-Target' => 'DataPipeline.CreatePipeline' },
           })
-
-          Fog::JSON.decode(response.body)
```
geemus: Ah. I'm open to either way. Not major-bumping is obviously easier. I think this stuff is relatively new, so that makes it feel a bit more flexible to change. But I agree it could break things. What is your inclination?

lanej: My inclination is release a major bump and be good semver people.

geemus: Works for me, I think. My only concern would be about how weird it might be (or not) for fog-core to be in a different major release level. They are distinct projects now, so probably fine, but just something to note.

PR #323, opened 2016-11-25
Fix the bug that can't create fifo queue in SQS.

ebihara99999:
# About the bug
Trying to create FIFO queue following [the document](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html), the error occurs (ommited some lines):
```
/path/to/vendor/bundle/ruby/2.3.0/gems/excon-0.54.0/lib/excon/middlewares/expects.rb:7:in `response_call': Expected(200) <=> Actual(400 Bad Request) (Excon::Error::BadRequest)

InvalidParameterValue Can only include alphanumeric characters, hyphens, or underscores. 1 to 80 in length
```

The script is
```ruby
require 'fog/aws'

sqs = Fog::AWS::SQS.new region: "us-west-2"
sqs.create_queue(name="FifoQueue.fifo", {"Attribute.1.Name" => "FifoQueue" , "Attribute.1.Value" => "true"})
```

# How to fix
The error says a queue name can't include '.' but fifo queue needs the '.fifo' suffix.
The version of query api fog-awe uses in SQS is behind from the latest and doesn't support the fifo queue (or more exactly, the old version doesn't support a queue name including '.', which doesn't necessarily mean whether it supports FIFO queue).
I changed the version from `2009-02-01` into `2012-11-05`, and the error doesn't show up.

# Testing
Executed `bundle exec shindo tests/requests/sqs`. Results are all green.

### DIFF
--- a/lib/fog/aws/sqs.rb
+++ b/lib/fog/aws/sqs.rb
@@ -128,7 +128,7 @@ module Fog
               :host               => @host,
               :path               => path || @path,
               :port               => @port,
-              :version            => '2009-02-01'
+              :version            => '2012-11-05'
             }
           )
 


ebihara99999:
Considering mocks, it seems I should've created kind of `#create_fifo_queue`....
You can see the `"FifoQueue": "true"` attribute when get api. 
Or in mock, I should've added conditional statements judging whether it's a fifo queue.
I would appreciate if someone would comment.

lanej:
@ebihara99999 i can't speak to your specific fifo queue issue but I can confirm that upgrading the version does work correctly.  Could you link a specific references of `#create_fifo_queue` ?

ebihara99999:
Sorry for a confusing report and comment.

Let me rephrase the bug:
* A FIFO queue name needs to have the .fifo suffix.
* The old version doesn't support a queue name including '.' (as mentioned above).
* I judged it is a bug in creating a FIFO queue, not a bug of depending on an old version api.

And I added the comment, if it's a bug about FIFO queue, because
* I think I should've modified `Mock.create_queue`.
* But the api sets the FifoQueue attribute as follows:
  * When it's a FIFO queue, FifoQueue is true.
  * When it's NOT a FIFO queue, the FifoQueue attribute doesn't exist in the attribute. Confirming with AWS CLI,

  ```shell
  $ SQS_QUEUE_URL="URL of NOT a fifo queue"

  $ aws sqs get-queue-attributes --queue-url ${SQS_QUEUE_URL} --attribute-names FifoQueue
  An error occurred (InvalidAttributeName) when calling the GetQueueAttributes operation: Unknown Attribute FifoQueue.
  ```

  ```shell
  $ SQS_QUEUE_URL="URL of a FIFO queue"
  $ aws sqs get-queue-attributes --queue-url ${SQS_QUEUE_URL} --attribute-names FifoQueue
  {
      "Attributes": {
          "FifoQueue": "true"
      }
  }
  ```
* Considering this, I have no idea how to integrate the FifoQueue attribute into the existing `Mock.create_queue`, and I left the comment (Creating '#create_fifo_queue' would solve the problem).

lanej:
@ebihara99999 thanks for the clarification.  This section of the codebase has no object modeling and the request options are freeform.  You would create a fifo queue by calling:

```ruby
sqs.create_queue 'queue.fifo', 'FifoQueue' => true
```

Bumping the API version is definitely required.

PR #327, opened 2016-12-05
fix host header with another port on s3

rodrigoapereira:
I'm using minio server on development environment, and catch this case using pre-signed url.

But i'm not find way to test this case.


geemus:
Thanks!

PR #329, opened 2016-12-12
credential_fetcher: Mark AWS metadata calls as idempotent

mtekel:
We would like to handle temporary issues with credential retrieval from AWS metadata server (usually running at 169.254.169.254) in a bit more resilient way. One possibility to achieve this is to mark the calls as idempotent, which means excon will retry them up to 3 times.

We are proposing the changes to the code in this PR. Please let us know if you have any comments or suggestions for improvement.


lanej:
Thanks @mtekel

geemus:
Thanks!

PR #332, opened 2016-12-15
Updated ELB Dual Stack hosted zone DNS records

mattheworiordan:
There are many new regions, this is the complete list as of today.  Very annoying that we should have to maintain this arbitrary mapping, and annoying that Amazon states in their docs it is auto-assigned in the AWS console yet not in the API.


geemus:
Yeah, definitely annoying... Thanks for the fix!

PR #336, opened 2017-01-04
Fix parsing of the Reserved Instance 'recurringCharge' field and add 'scope' field

sodabrew:
To get access to the new Region reserved instances, I needed to add 'scope' to the parser, and specify a newer API version when creating the AWS connection, i.e.

```
AWS = Fog::Compute.new({
  provider: 'AWS',
  version: '2016-11-15',
  ...
})
```

### DIFF
--- a/lib/fog/aws/parsers/compute/describe_reserved_instances.rb
+++ b/lib/fog/aws/parsers/compute/describe_reserved_instances.rb
@@ -25,7 +25,7 @@ module Fog
 
           def end_element(name)
             case name
-            when 'availabilityZone', 'instanceTenancy', 'instanceType', 'offeringType', 'productDescription', 'reservedInstancesId', 'state'
+            when 'availabilityZone', 'instanceTenancy', 'instanceType', 'offeringType', 'productDescription', 'reservedInstancesId', 'scope', 'state'
               @reserved_instance[name] = value
             when 'duration', 'instanceCount'
               @reserved_instance[name] = value.to_i
--- a/lib/fog/aws/parsers/compute/describe_reserved_instances.rb
+++ b/lib/fog/aws/parsers/compute/describe_reserved_instances.rb
@@ -4,16 +4,15 @@ module Fog
       module AWS
         class DescribeReservedInstances < Fog::Parsers::Base
           def get_default_item
-            {'tagSet' => {}, 'recurringCharges' => {}}
+            {'tagSet' => {}, 'recurringCharges' => []}
           end
 
           def reset
             @context = []
-            # Note:  <recurringCharges> should also be handled as a set, but do not want to disrupt anyone relying on
-            # it currently being flatted
-            @contexts = ['reservedInstancesSet', 'tagSet']
+            @contexts = ['reservedInstancesSet', 'recurringCharges', 'tagSet']
             @reserved_instance = get_default_item
             @response = { 'reservedInstancesSet' => [] }
+            @charge = {}
             @tag = {}
           end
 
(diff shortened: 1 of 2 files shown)


sodabrew:
Digging further, not all of the reservations on my account have a 'scope' parameter. I'm trying to figure out what's different about reservations with a specific scope and without...

geemus:
It's also probably fine to just add this when it is available maybe? Do let me know if you figure it out and/or if you think we should just bring this in as-is. Thanks!

sodabrew:
I left off yesterday dumping the raw output from the API response, but haven't sorted through the output yet. I'll try to do that today.

Interestingly, if I provided an earlier API version, the Region reservation wasn't returned at all. <-- useful note to future reserved instance users :)

sodabrew:
There's a bug in this parser, since, a long time. Sometimes my parsed results come out like this, with a second array holding an empty tagSet and recurringCharges field. I've seen this bug for over a year, but I assumed it was an AWS quirk and worked around it. What I'm seeing now is that the `scope` field is showing up in the second hash when it is missing from the first hash.

```
{"tagSet"=>{},
  "recurringCharges"=>{},
  "reservedInstancesId"=>"xxx",
  "instanceType"=>"xxx",
  "availabilityZone"=>"us-east-1e",
  "start"=>xxx,
  "end"=>xxx,
  "duration"=>31536000,
  "fixedPrice"=>xxx,
  "usagePrice"=>0.0,
  "instanceCount"=>xxx,
  "productDescription"=>"Linux/UNIX (Amazon VPC)",
  "state"=>"active",
  "instanceTenancy"=>"default",
  "offeringType"=>"Partial Upfront",
  "amount"=>xxx},
 {"tagSet"=>{}, "recurringCharges"=>{}, "scope"=>"Availability Zone"},
```

Here's how some of my other reservations look:

```
{"tagSet"=>{},
  "recurringCharges"=>{},
  "reservedInstancesId"=>"xxx",
  "instanceType"=>"xxx",
  "availabilityZone"=>"us-east-1e",
  "start"=>xxx,
  "end"=>xxx,
  "duration"=>31536000,
  "fixedPrice"=>xxx,
  "usagePrice"=>0.0,
  "instanceCount"=>xxx,
  "productDescription"=>"Linux/UNIX (Amazon VPC)",
  "state"=>"active",
  "instanceTenancy"=>"default",
  "offeringType"=>"All Upfront",
  "scope"=>"Availability Zone"},
```

The consistent theme is that `offeringType => Partial Upfront` reservations fail to parse. This is probably because the Partial Upfront reservations have a nested `recurringCharges` that *fails* to parse.

Here's what a `recurringCharges` raw XML looks like:

```
            <offeringType>Partial Upfront</offeringType>
            <recurringCharges>
                <item>
                    <frequency>Hourly</frequency>
                    <amount>0.03135</amount>
                </item>
            </recurringCharges>
```

The problem appears to be in the parser code, it knows what to do with an `item` tag, but isn't prepared to handle an `item` tag that's nested inside:

```
            when 'item'
              case @context.last
              when 'reservedInstancesSet'
                @response['reservedInstancesSet'] << @reserved_instance
                @reserved_instance = get_default_item
              when 'tagSet'
                @reserved_instance['tagSet'][@tag['key']] = @tag['value']
                @tag = {}
              end
```

So I think the solution is to push a new context for `recurringCharges`. This old NOTE indicates someone looked at this at some point?

```
            @context = []
            # Note:  <recurringCharges> should also be handled as a set, but do not want to disrupt anyone relying on
            # it currently being flatted
            @contexts = ['reservedInstancesSet', 'tagSet']
```

lib/fog/aws/parsers/compute/describe_reserved_instances.rb
```diff
@@ -4,16 +4,15 @@ module Compute
...
         class DescribeReservedInstances < Fog::Parsers::Base
           def get_default_item
-            {'tagSet' => {}, 'recurringCharges' => {}}
+            {'tagSet' => {}, 'recurringCharges' => []}
           end
 
           def reset
             @context = []
-            # Note:  <recurringCharges> should also be handled as a set, but do not want to disrupt anyone relying on
-            # it currently being flatted
```
sodabrew: I'm pretty sure this is false, nobody can be relying on this because it never worked :)

sodabrew:
@geemus This PR is ready for another look.

geemus:
Great, thanks for the detailed analysis and fix!

PR #339, opened 2017-01-19
Better iam policies

ehowe:


geemus:
Seems good to me from a high level. Defer to @lanej for final review/merge.

lib/fog/aws/models/iam/managed_policies.rb
```diff
@@ -51,6 +54,15 @@ def all_by_group(group_name, options={})
           end
         end
 
+        def all_by_role(role_name, options={})
+          body = service.list_attached_role_policies(role_name, page_params(options)).body
+          merge_attributes(body)
+
+          body['Policies'].map do |policy|
```
lanej: This is a frequently duplicated block.  Should extract this at some point.

lanej:
Thanks @ehowe

PR #340, opened 2017-01-21
Fix multipart upload

nobmurakita:
Add missing 'self' to call setter method

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -210,7 +210,7 @@ module Fog
           options.merge!(encryption_headers)
 
           # With a single PUT operation you can upload objects up to 5 GB in size. Automatically set MP for larger objects.
-          multipart_chunk_size=5242880 if !multipart_chunk_size && Fog::Storage.get_body_size(body) > 5368709120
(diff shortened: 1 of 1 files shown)


geemus:
I think the self should be implied there, and I don't believe I've heard about any related issues. Were you having a particular error that this fixes? Just trying to make sure I understand the context. Thanks!

nobmurakita:
If `self` is missing, the local variable `multipart_chunk_size` is created and can't access `@multipart_chunk_size`.
When body size is smaller than 5 GB, `multipart_chunk_size` on the following code is `nil`.
Multipart upload doesn't work even if `@multipart_chunk_size` is set.

```
if multipart_chunk_size && Fog::Storage.get_body_size(body) >= multipart_chunk_size && body.respond_to?(:read)
  data = multipart_save(options)
  merge_attributes(data.body)
else
```

lanej:
Small verification of @nobmurakita's claim.

```ruby
Class.new do
  attr_accessor :foo
  def bar
       foo = 'bar'
    end
end.new.tap { |x| x.bar }.foo == 'bar' # false
```

@geemus The issue and the fix are valid but I'll leave it you to confirm.

geemus:
@nobmurakita thanks for the clarification.

@lanej thanks for the verification.

PR #346, opened 2017-02-15
add natGatewayId to describe_route_tables

mliao2:
I would like to see natGatewayId when I do describe_route_tables.


geemus:
Thanks!

PR #350, opened 2017-03-06
Modify volume

ehowe:


geemus:
Looks like there is a syntax error related to 1.8, which we should kill anyway probably, but fwiw. Otherwise looks good to me. @lanej will leave for you to double check. Thanks!

ehowe:
@geemus it would be great if we could kill 1.8 support sometime.  Either way, that syntax error should be fixed, thanks for pointing it out.

geemus:
Agreed, we need to get rid of it. But for now, if they are small changes we should I guess make them just a little longer.

lib/fog/aws/models/compute/volume.rb
```diff
@@ -36,31 +36,50 @@ def ready?
           state == 'available'
         end
 
-        def save
-          raise Fog::Errors::Error.new('Resaving an existing object may create a duplicate') if persisted?
```
lanej: This case is no longer applicable?

lib/fog/aws/requests/compute/modify_volume.rb
```diff
@@ -63,7 +63,7 @@ def modify_volume(volume_id, options={})
           if options['Iops']
             volume_modification.merge!(
               'originalIops' => volume['iops'],
-              'targetIops'   => options['Iops'],
+              'targetIops'   => options['Iops']
```
lanej: how dare you eugene.

lib/fog/aws/models/compute/volume.rb
ehowe: No, since if there is an identity, its going to call the modify_volume method instead of create_volume.

lib/fog/aws/requests/compute/modify_volume.rb
ehowe: I know, right?

geemus:
Thanks!

PR #351, opened 2017-03-09
Add check for self.etag before running gsub

dmcorboy:
Matches to https://github.com/fog/fog/blob/master/lib/fog/internet_archive/models/storage/file.rb


lanej:
Thanks @dmcorboy

geemus:
Thanks!

PR #353, opened 2017-03-09
Add new i3 class instances.

rogersd:


lanej:
From https://aws.amazon.com/ec2/instance-types/

image

lib/fog/aws/models/compute/flavors.rb
```diff
@@ -425,6 +425,66 @@ class AWS
...
+          :ebs_optimized_available => true,
+          :instance_store_volumes  => 4
+        },
+        {
+          :id                      => 'i3.16xlarge',
+          :name                    => 'I3 Sixteen Extra Large',
+          :bits                    => 64,
+          :cores                   => 32,
+          :disk                    => 15200,
+          :ram                     => 249856,
```
lanej: This should be `(488 * 1024) = 499712`.

rogersd: Curse you, copy and paste!

rogersd:
Fixed.

geemus:
Thanks!

PR #358, opened 2017-03-30
Update CHANGELOG for v1.3.0

tricknotes:
These features are released as v1.3.0 :tada:


lanej:
oops. I updated it before I saw this but good looking out.

PR #360, opened 2017-04-20
Skip call to instance metadata host if region is specified

ankane:
Hey, thanks for the awesome work on this gem 👍 

This PR avoids an extra fetch if region is specified.

The main benefit isn't performance, but rather permissions. For containers, AWS recommends blocking the instance metadata host (with `iptables`) so containers running on Amazon ECS don't have access to the underlying EC2 instance credentials.

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html


geemus:
Thanks!

PR #362, opened 2017-04-26
Add p2 instance types

Caged:
This adds the [p2 instance types](https://aws.amazon.com/ec2/instance-types/p2/).  These are gpu-based, but there wasn't a key for specifying the number of GPUs, so I followed the existing pattern.


geemus:
@Caged we started with these before gpu-based were available, so I think the lack of a particular value here can be chalked up to historical coincidence. Perhaps we should revisit that and think about adding the information now that it is needed. For now though, this looks great, thanks!

PR #366, opened 2017-05-23
Fixed credential refresh when instance metadata host is inaccessible

ankane:
Pass `region` to credential refresh if specified. Otherwise, Fog tries to call instance metadata host. More context in #360.


geemus:
Thanks!

PR #367, opened 2017-05-24
Add a top-level require that matches the gem name

lanej:
Reduce confusion by allowing the gem to be required via a file that
matches the gem name.

References https://github.com/fog/fog/issues/3959


geemus:
👍  LGTM, thanks!

Issue #369, opened 2017-06-08
Support REST Bucket Get v2


khoan:
Parser doesn't support ContinuationToken as per http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html

```
require 'fog/aws'

Fog::Storage.new(
  :provider => 'AWS',
  :aws_access_key_id => ':key',
  :aws_secret_access_key => ':secret'
).tap do |s|
  r = s.get_bucket(':bucket', 'list-type' => 2)

  if r.body['IsTruncated']
    puts r.body['ContinuationToken'].inspect # => nil
  end
end
```

returns expected result after monkey patch

```
require 'fog/aws/parsers/storage/get_bucket'
Fog::Parsers::Storage::AWS::GetBucket.class_eval do
  alias_method :original_end_element, :end_element

  def end_element name
    case name
    when 'NextContinuationToken'
      @response['NextContinuationToken'] = value
    else
      original_end_element name
    end
  end
end
```

geemus:
Thanks for the detailed report. Would you be up for making the monkey patch into a quick PR for us so we can get that fixed? Thanks!

PR #370, opened 2017-06-14
fix https://github.com/fog/fog-aws/issues/369

khoan:
not sure how to test this, so no tests were added.


geemus:
Thanks, it's a pretty trivial fix so I'm not super worried test-wise. Worst case it still wouldn't work, but I don't think it should break anything. Thanks!

PR #374, opened 2017-06-30
Fix AWS credential mocking

v-yarotsky:
The change introduced in #252 does not actually prevent HTTP requests
from being made because of a missing early return statement.


geemus:
👍  LGTM, thanks!

geemus:
@lanej there was desire for this to be in a released version. Would you be able to cut one? Just let me know if you don't have availability and I can try to find some time also. Thanks!

lanej:
https://rubygems.org/gems/fog-aws/versions/1.4.1

geemus:
@lanej awesome, thanks!

PR #376, opened 2017-07-06
Add MaxResults filter to describe reserved instances offerings

KevinLoiseau:
A little PR for :
- Add `MaxResults` and `NextToken` filter to describe_reserved_instances_offerings.
- Update the serializer for adding `nextToken` in the result payload.


geemus:
Thanks!

PR #381, opened 2017-08-02
added support to retrieve and create vpc with ipv6 cidr block

chanakyacool:


geemus:
LGTM.

@lanej - could you double check for us and let me know if you have any concerns? Thanks!

geemus:
Thanks!

PR #382, opened 2017-08-04
Allow specifying kms key id to use

fcheung:
The file model currently only supports using the default kms key, this adds support for passing a specific key id to use.


geemus:
LGTM. Thanks!

PR #386, opened 2017-08-31
Ability to save cidrBlockAssociationSet

davidv-cuelogic:
describe_vpcs now return Ipv6CidrBlockAssociationSet and CidrBlockAssociationSet as a list of association objects. 
This PR attempts to save both the results in ipv_6_cidr_block_association_set and cidrBlockAssociationSet attributes respectively

### DIFF
--- a/lib/fog/aws/models/compute/vpc.rb
+++ b/lib/fog/aws/models/compute/vpc.rb
@@ -11,6 +11,7 @@ module Fog
         attribute :tenancy,          :aliases => 'instanceTenancy'
         attribute :is_default,       :aliases => 'isDefault'
         attribute :ipv_6_cidr_block_association_set,  :aliases => 'ipv6CidrBlockAssociationSet'
+        attribute :cidrBlockAssociationSet,  :aliases => 'cidrBlockAssociationSet'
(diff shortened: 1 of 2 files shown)


lib/fog/aws/models/compute/vpc.rb
```diff
@@ -11,6 +11,7 @@ class VPC < Fog::Model
         attribute :tenancy,          :aliases => 'instanceTenancy'
         attribute :is_default,       :aliases => 'isDefault'
         attribute :ipv_6_cidr_block_association_set,  :aliases => 'ipv6CidrBlockAssociationSet'
+        attribute :cidrBlockAssociationSet,  :aliases => 'cidrBlockAssociationSet'
```
geemus: For consistency I think we probably want the attribute name to be underscore separated instead of camelcase?

lib/fog/aws/parsers/compute/describe_vpcs.rb
```diff
@@ -24,22 +28,33 @@ def end_element(name)
               case name
                 when 'item'
                   @vpc['tagSet'][@tag['key']] = @tag['value']
-                  @tag = {}
```
geemus: Why is this removed? Seems like it would be unrelated to the cidr block sets and might break stuff.

lanej:
This is covered in #389

PR #391, opened 2017-09-14
Add ability to encrypt a copy of an unencrypted snapshot

nodecarter:
This PR will allow to use `Encrypted` and `KmsKeyId` parameters while copying snapshots.
In other words will allow to encrypt a copy of an unencrypted snapshot.

http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CopySnapshot.html

### DIFF
--- a/lib/fog/aws/requests/compute/copy_snapshot.rb
+++ b/lib/fog/aws/requests/compute/copy_snapshot.rb
@@ -9,8 +9,8 @@ module Fog
         # ==== Parameters
         # * source_snapshot_id<~String> - Id of snapshot
         # * source_region<~String>      - Region to move it from
-        # * description<~String>        - A description for the EBS snapshot
         # * options<~Hash>:
+        #   * 'Description'<~String>    - A description for the EBS snapshot
         #   * 'Encrypted'<~Boolean>     - Specifies whether the destination snapshot should be encrypted
         #   * 'KmsKeyId'<~String>       - The full ARN of the AWS Key Management Service (AWS KMS) CMK
         #                                 to use when creating the snapshot copy.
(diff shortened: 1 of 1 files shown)


lib/fog/aws/requests/compute/copy_snapshot.rb
```diff
@@ -17,14 +22,17 @@ class Real
         #     * 'snapshotId'<~String> - id of snapshot
         #
         # {Amazon API Reference}[http://docs.amazonwebservices.com/AWSEC2/latest/APIReference/ApiReference-query-CopySnapshot.html]
-        def copy_snapshot(source_snapshot_id, source_region, description = nil)
-          request(
+        def copy_snapshot(source_snapshot_id, source_region, description = nil, options = {})
```
geemus: It seems a little weird to me to have the defaulting-to-nil description and then options afterward, instead I wonder if just trying to pull `description` from options (where it will be nil by default) would be a more consistent interface?

I realize this wouldn't be directly backwards compatible, but I think we can fake it (pretty sure we've done this elsewhere). Basically you just expect 3 arguments, and you check to see if the 3rd one is a hash or string. If it is a hash you just do what your code already shows, but if it is a string you convert it into the hash form and output a deprecation warning (to train people to update their usage without breaking anything).

I realize this adds time and effort to what seems like it should be a pretty simple change, but I think it might be the better interface in the longer term. What do you think? Would you be able to help us with that revision? Thanks!

nodecarter: Yes, I agree, it make sense.

geemus:
Thanks!

Issue #387, opened 2017-09-05
Fog::Compute::AWS::Vpcs returns VPCs with nil ids


maguec:
Ruby Version: 2.3.3 and 2.4.1
Fog Versions:
fog-aws (1.4.0)
fog-core (1.45.0)
fog-json (1.0.2)
fog-xml (0.1.3)

Repro case:

```
$ irb                                                                                                                                                                        
2.3.3 :001 > require 'fog-aws'                                                                                                                                               
 => true                                                                                                                                                                     
2.3.3 :002 > conf = { provider: 'AWS', region: 'us-west-1', aws_access_key_id: ENV['AWS_ACCESS_KEY_ID'], aws_secret_access_key: ENV['AWS_SECRET_ACCESS_KEY'] }      
<SNIP>
2.3.3 :003 > new_conn = Fog::Compute.new(conf)           
2.3.3 :004 > new_conn.vpcs.all
 =>   <Fog::Compute::AWS::Vpcs
    filters={}
    [
                  <Fog::Compute::AWS::VPC
        id="vpc-blahblah",
        state="associated",
        cidr_block="192.168.1.0/24",
        dhcp_options_id="default",
        tags={},
        tenancy="default",
        is_default=nil
      >,
                  <Fog::Compute::AWS::VPC
        id=nil,
        state=nil,
        cidr_block=nil,
        dhcp_options_id="dopt-78b8761c",
        tags={"Name"=>"MYVPC"},
        tenancy="default",
        is_default=false
      >,
                  <Fog::Compute::AWS::VPC
        id="vpc-blahblah",
        state="associated",
        cidr_block="172.31.0.0/16",
        dhcp_options_id="default",
        tags={},
        tenancy="default",
        is_default=nil
      >,
                  <Fog::Compute::AWS::VPC
        id=nil,
        state=nil,
        cidr_block=nil,
        dhcp_options_id="dopt-00ac6264",
        tags={},
        tenancy="default",
        is_default=true
      >
    ]
  >
```

Compared to the aws CLI

```
$ aws ec2 describe-vpcs --region us-west-1
{
    "Vpcs": [
        {
            "Tags": [
                {
                    "Key": "Name",
                    "Value": "MYVPC"
                }
            ],
            "InstanceTenancy": "default",
            "DhcpOptionsId": "dopt-78b8761c",
            "CidrBlock": "192.168.1.0/24",
            "State": "available",
            "VpcId": "vpc-blahblah",
            "IsDefault": false
        },
[...]

PR #396, opened 2017-09-20
Issue #387 Fog::Compute::AWS::Vpcs returns VPCs with nil ids

maguec:
handle the case where item ends up being "associated" instead of a staring beginning with true or false


geemus:
Thanks!

PR #397, opened 2017-09-22
Fix detaching instances from auto scaling group.

rzaharenkov:
React request works as expected. This change only affects mock mode.


rzaharenkov:
Guys, it's really tiny bit of code required for specs. Could you please review it?

Issue #335, opened 2016-12-21
Support for AWS Application Load Balancer (ALB)


daugustus:
I am unable to find any details about using Fog with the new Application Load Balancers that AWS. Any news on this?

https://aws.amazon.com/blogs/aws/new-aws-application-load-balancer/

Thanks,
Dave

geemus:
Things are generally added as somebody has a need (or just desire) to use them. I haven't heard any other discussion, so you may be the first one to raise it so far. Would you be interested in helping us work toward that? Thanks!

daugustus:
That was a thought I had as well. I will take a look and see what is involved.

geemus:
Sounds good, just let us know if there is anything we can assist with.

daugustus:
To get started on this, do you have any pointers or guides for this project? or for Ruby in general?

geemus:
There is a high level overview here: https://github.com/fog/fog/blob/master/CONTRIBUTING.md as well as http://fog.io/about/structure.html. If you are familiar with one or more of the existing aws services, a lot should be pretty similar too. As for Ruby, there are many good resources online, but I'm a bit of an old hand at this point so I'm not sure I have good recommendations for someone more new to it. Hope that helps!

rockhunters08:
Has anyone started to work on this?  I could use it, so would do so if not.

lanej:
@rockhunters08 afaik no one has started this and we'd be happy to receive a PR adding support for it.

rockhunters08:
@lanej Ok, I'll start taking a look at it next week. Thanks.

tomazstrukeljdlabs:
Does method target_health_descriptions work yet?
- target_health_descriptions returns "(Object doesn't support #inspect)"
- target_health_descriptions.get(ARN) returns "NoMethodError: undefined method `id' for nil:NilClass"

PR #400, opened 2017-09-26
Initial support for ALB

starbeast:
Models, requests and services to work with application load balancers.
Necessity of the feature was described and discussed in #335


geemus:
Looks like this may be somewhat of a work in progress still. Just let us know when it is ready for review or if you have questions we can help with. Thanks!

lib/fog/aws/parsers/elbv2/describe_target_health.rb
```diff
@@ -0,0 +1,56 @@
...
+            when 'Description', 'Reason', 'State'
+              @target_health[name] = value if @in_target_health
+
+            when 'Target'
+              @target_health_description[name] = @target
+
+            when 'TargetHealth'
+              @target_health_description[name] = @target_health
+
+            when 'Targets'
```
rockhunters08: not sure the right fix, but I needed to change 'Targets' above to 'TargetHealthDescriptions' in order to get the actual data from the request bubbled up.

spatino:
Hi, is there any news about this old PR to be merged in master?

PR #405, opened 2017-10-19
Drop Ruby<2.0 support

tbrisker:
Users requiring ruby 1.9 or older should use the main fog gem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,6 @@ branches:
 matrix:
   fast_finish: true
   include:
-  - rvm: 1.8.7
-    gemfile: gemfiles/Gemfile-ruby-1.8.7
-  - rvm: 1.9.3
-    gemfile: gemfiles/Gemfile-ruby-1.9
   - rvm: 2.0.0
     gemfile: gemfiles/Gemfile-ruby-2.0
   - rvm: 2.1.0
@@ -27,8 +23,6 @@ matrix:
     gemfile: gemfiles/Gemfile-edge
   - rvm: 2.3.0
     gemfile: Gemfile
-  - rvm: jruby-19mode
-    gemfile: gemfiles/Gemfile-ruby-1.9
   - rvm: jruby-head
     gemfile: Gemfile
   allow_failures:
--- a/fog-aws.gemspec
+++ b/fog-aws.gemspec
@@ -19,6 +19,8 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ["lib"]
 
+  spec.required_ruby_version = '>= 2.0.0'
+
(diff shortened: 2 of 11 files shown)


geemus:
@tbrisker I think maybe the main fog gem requires this though, so it would force the dependency change on it as well? (which we should probably do anyway, but that we have so far treated as a breaking change). Does that make sense? Thoughts?

tbrisker:
@geemus do you mean `fog/fog` or `fog/fog-core`? From what I understand from the documentation, `fog/fog` should maintain legacy support while `fog/fog-core` should only support ruby >= 2. I don't mind opening a PR to `fog/fog-core` to clarify this and remove some leftovers there as well.
As far as this gem is concerned, I would say that at least a minor version bump makes sense, if not a major one.

plribeiro3000:
@geemus Some other providers already accepted and merged this change.
I believe it might be okay since we have a version constraint the the gemspec so we wont have any breakeages, old versions of ruby will just not receive more updates. They will be locked in the last current version.

I know this is bad but we need to catch up, we are already 3 versions behind. At some point we wont be able to maintain this train anymore. 😭 

Maybe we should let it go now and if a change is really need for older ruby versions we can backport it.

WDYT?

geemus:
@tbrisker I had meant `fog/fog`. My memory on some of these things can be fuzzy though...

@plribeiro3000 ah. Thanks for the context.

So if I read all that correctly (and please one or both of you correct me if not). We should add a lock to the current version of `fog-aws` to `fog/fog` in addition to this change and then we will be good to go? Or you are saying that the version constraint here will mean that anyone with less than 2.0 will just automatically get the newest version without that constraint? (apologies, I just haven't really made use of the version constraint like this much before).

plribeiro3000:
Yeap. With this constraint, `bundler` wont allow anyone with `ruby` prior to `2.0` to install new versions.

It means that we might get some issues of people using old and perhaps buggy versions of `fog`.
And we have to ask them to move on to ruby `> 2.0` and preferable `2.3`.

tbrisker:
Looks like `fog/fog` itself has actually dropped 1.9 support about 2 years ago in https://github.com/fog/fog/commit/8a413ef54a82e902c59e1e277a38ffe91a466968 - so I think we're good.
And considering that 1.9 is EOL for almost 3 years already, I would say it is reasonable to tell anyone still on it to either use older versions of the fog gem or upgrade to more modern ruby (even 2.1 is already EOL).

tbrisker:
I think perhaps it's time to actually cut a fog 2.0 version that drops support for legacy ruby and lock the 1.x branch on last supporting versions, and only backport serious security fixes to the 1.x branch and relevant gems.

plribeiro3000:
If we are gonna do that, perhaps we should do it for all EOL ruby versions.

geemus:
Bringing this in for now, we can certainly still discuss further changes though. Thanks!

Issue #371, opened 2017-06-20
Not all Rds versions and Instance Types are rendered


chanakyad-cuelogic:
ref: `fog/aws/requests/rds/describe_orderable_db_instance_options.rb`
- When compared to AWS Console many versions were found missing 
For Eg:-
Scenario considered for Demonstration  
 - region:- 'ap-south-1'
 - engine:- 'postgres'
1. Through API 
- `c.describe_orderable_db_instance_options('postgres').data[:body]["DescribeOrderableDBInstanceOptionsResult"]["OrderableDBInstanceOptions"]`
=> Uniq versions were found ["9.4.7", "9.4.9", "9.4.11"]
2. When on AWS console and AWS API Reference (DescribeOrderableDBInstanceOptions)
 refer screenshot attached 
![screen shot 2017-06-20 at 8 32 14 pm](https://user-images.githubusercontent.com/8611339/27340415-04ad8636-55f8-11e7-8462-e2e2a74efa87.png)

geemus:
We may need to bump the version of RDS we are using to get all the options? Otherwise I'm not sure why there would be such discrepancies.

chanakyad-cuelogic:
@geemus 
will look at the version of rds to get all the options, i hope that would assist me to solve this 
Thanks 👍

carloslima:
@chanakyad-cuelogic

You don't see all options because of the limit (100 maximum), if there are more than that amount of items (which is the case for that request[1]), it is required to make another request to fetch the next batch, until you reach the end.

You can get all options with:
```
require 'fog/aws'
options=Fog::AWS::RDS.new(region: 'ap-south-1').instance_options(engine: 'postgres')
all_options = options.to_a
all_options.concat(options.to_a) while options.filters[:marker] and options.all

puts all_options.map(&:engine_version).uniq
puts all_options.size

# 9.4.7
# 9.4.9
# 9.4.11
# 9.4.12
# 9.5.2
# 9.5.4
# 9.5.6
# 9.5.7
# 9.6.1
# 9.6.2
# 9.6.3
# 462
```

[1]: Just for clarity, each version is returned multiple times for each combination of instance size and options.

@geemus

In the code, Fog::AWS::RDS::InstanceOptions is a PagedCollection and it's implementation of #all has a comment stating that it intentionally only fetches the first page.

This makes me confused because considering there are other cases of paged collections, I would expect a standard way of:
1. knowing you got truncated results
2. a way to request the next page
3. a way to request all pages

[...]

PR #406, opened 2017-10-22
Update rds instance options model

carloslima:
While looking at issue #371 I noticed that the AWS API returns more data than was being exposed by fog, so I updated the model and parser to expose the remaining items and fixed a mismatch on checking for IOPS Support.

### DIFF
--- a/lib/fog/aws/models/rds/instance_option.rb
+++ b/lib/fog/aws/models/rds/instance_option.rb
@@ -10,6 +10,12 @@ module Fog
         attribute :availability_zones, :aliases => 'AvailabilityZones', :type => :array
         attribute :db_instance_class, :aliases => 'DBInstanceClass'
         attribute :vpc, :aliases => 'Vpc', :type => :boolean
+        attribute :supports_iops, :aliases => 'SupportsIops', :type => :boolean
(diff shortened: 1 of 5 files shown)


geemus:
@carloslima looks reasonable to me.

@lanej could you review as well? Thanks!

lanej:
@carloslima Could you modify https://github.com/fog/fog-aws/blob/650a5884c70c31b647bc7933dc402c59aef14dde/lib/fog/aws/requests/rds/describe_orderable_db_instance_options.rb#L7-L8

to link to http://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_OrderableDBInstanceOption.html.

Looks good otherwise.

carloslima:
Nice, thanks for catching that.

I've used this link instead:
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeOrderableDBInstanceOptions.html

The one you sent is for the response.
Should I add it to the parser? I didn't see docs on other parsers.

lanej:
@carloslima Good point.  Don't worry about the parser, it's easy to correlate to the request.

geemus:
Thanks!

PR #407, opened 2017-10-23
Add emulation of default VPC to mocked mode.

rzaharenkov:
When you create some resources in VPC-enabled environment (security groups, servers, load balaners etc.) AWS automatically attaches it to the default VPC. I am trying to emulate the same behavior in mocked mode. 

In order to to it I am going to create following resources:
- default vpc;
- internet gateway;
- routes table;
- subnets for each availability zone.

It's not possible to create such resources using requests cause:

1. It doesn't accept some attributes (e.g. `is_default` or `default_for_az`).
2. It creates resources in :pending state.

So I am going to push raw data to `data` collection.

Also I don't think it should be created by default because it may break some existing specs. So I propose to add a new method `Fog::Compute::AWS::Mock#setup_default_vpc!`. All data is stored in the class attribute so in order to use it you can create an instance of compute service and call this method, e.g.:

```ruby
compute = Fog::Compute.new(provider: 'aws', region: 'us-east-1')
compute.setup_default_vpc!
security_group = compute.create_security_group('test_group', 'New test group')
expect(security_group.vpc_id).to be_present
```

I am looking forward for your comments/suggestions.

### DIFF
--- a/lib/fog/aws/compute.rb
+++ b/lib/fog/aws/compute.rb
@@ -196,6 +196,8 @@ module Fog
           'vpc'  => 'vpc'
         }
 
+        VPC_BLANK_VALUE = 'none'
+
         include Fog::AWS::CredentialFetcher::ConnectionMethods
 
         def self.data
@@ -284,7 +286,7 @@ module Fog
                     "attributeName" => "supported-platforms"
                   },
                   {
-                    "values"        => ["none"],
+                    "values"        => [VPC_BLANK_VALUE],
                     "attributeName" => "default-vpc"
                   },
                   {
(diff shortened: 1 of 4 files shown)


geemus:
@rzaharenkov sounds pretty reasonable to me, thanks for the detailed PR. I would like to hear more about the specs that break if you do this by default. Do you think they really should break, or do they reflect things we should update as well?

@lanej could you also review? Thanks!

rzaharenkov:
@geemus , I am worried regarding some specs that may expect exact number of networks or subnets. That's why I prefer to turn it on manually whether it's needed.

geemus:
@rzaharenkov That makes sense, I guess it just made me wonder if those tests are also currently broken if/when you run against real AWS. If they are, then our tests may be broken to begin with. Regardless, testing certain counts there may be overly specific/brittle. Thanks for your insight, I'm ok bringing this in as-is for now but wanted to discuss.

rzaharenkov:
The thing is AWS works in the different way depend on creation date of the account and region. It depends on supported platforms property. You can see it on ec2 dashboard or in the response on `describe_account_attributes` requests. In case when you have both VPC and EC2 flags it won't create a default VPC.

Here is a screenshots of EC2 summary for the same account in different regions:

monosnap 2017-10-24 19-07-14
monosnap 2017-10-24 19-07-50

geemus:
Gross... That does make it much more complicated. Let's just bring this in as is, at least for now then. Thanks!

PR #408, opened 2017-10-25
Add jitter to exponential backoff

masstamike:
Using exponential backoff alone is not adequate when launching multiple instances simultaneously. Jitter added to compensate for this.

### DIFF
--- a/lib/fog/aws/compute.rb
+++ b/lib/fog/aws/compute.rb
@@ -560,7 +560,16 @@ module Fog
                   Fog::Compute::AWS::NotFound.slurp(error, match[:message])
                 when 'RequestLimitExceeded'
                   if retries < max_retries
-                    sleep (2.0 ** (1.0 + retries) * 100) / 1000.0
+                    jitter = rand(100)
+                    waiting = true
+                    start_time = Time.now
+                    wait_time = ((2.0 ** (1.0 + retries) * 100) + jitter) / 1000.0
+                    puts "Waiting #{wait_time} seconds to retry."
+                    while waiting
+                      if Time.now - start_time >= wait_time
+                        waiting = false
+                      end
+                    end
                     retries += 1
                     retry
                   else
--- a/lib/fog/aws/compute.rb
+++ b/lib/fog/aws/compute.rb
@@ -564,7 +564,7 @@ module Fog
                     waiting = true
                     start_time = Time.now
                     wait_time = ((2.0 ** (1.0 + retries) * 100) + jitter) / 1000.0
-                    puts "Waiting #{wait_time} seconds to retry."
+                    Fog::Logger.warning "Waiting #{wait_time} seconds to retry."
                     while waiting
                       if Time.now - start_time >= wait_time
                         waiting = false


.gitignore
```diff
@@ -14,3 +14,4 @@
 mkmf.log
 tests/.fog
 gemfiles/Gemfile-edge.lock
+.idea/
```
geemus: Could you omit this? I don't think it's strictly needed, so prefer to keep things streamlined.

lib/fog/aws/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fog
   module AWS
-    VERSION = "1.4.1"
+    VERSION = "1.4.2"
```
geemus: Also, if you could leave this out. We'll make this change when we cut a release.

geemus:
@masstamike thanks for bringing this to us. Is the jitter sufficient to make it all work consistently? Just makes me wonder if the backoff should be more aggressive instead? (ie I worry that if rand happened to pick a lot of low numbers it could still fail).

.gitignore
lanej: @masstamike You'll likely want to add `/.idea` to your global git ignores.  See https://help.github.com/articles/ignoring-files/

masstamike: Ah thanks @lanej! I'll remove that.

lib/fog/aws/compute.rb
```diff
@@ -560,7 +560,16 @@ def _request(body, headers, idempotent, parser, retries = 0)
                   Fog::Compute::AWS::NotFound.slurp(error, match[:message])
                 when 'RequestLimitExceeded'
                   if retries < max_retries
-                    sleep (2.0 ** (1.0 + retries) * 100) / 1000.0
+                    jitter = rand(100)
+                    waiting = true
+                    start_time = Time.now
+                    wait_time = ((2.0 ** (1.0 + retries) * 100) + jitter) / 1000.0
+                    puts "Waiting #{wait_time} seconds to retry."
```
geemus: We probably won't want a puts here. I think we should either omit or use `Fog::Logger` so that people can change levels/etc if desired, probably at debug level? What do you think?

geemus:
@masstamike I think it may have gotten lost in the shuffle, but I did wonder if the jitter change was sufficient. ie if rand was low numbers a few times in a row, would we still see failure? If so, should the falloff just be larger in general?

masstamike:
Hi @geemus, I was hoping to have some definitive testing done before answering that, but we haven't been able to fully test the new changes under heavy load quite yet, hopefully by end of day. But the problem we were facing would not be solved without jitter. The problem is 10 or more simultaneous `vagrant up` commands which hit the API limit then wait the same amount of time to try again, hitting the API limit once more, and continuing up to 9 times, before vagrant decides to not wait any longer. None of the processes even hit the 10 retry limit enforced here.

Another benefit to using jitter is decreased load on the server as referenced by [this AWS architecture blog](https://www.awsarchitectureblog.com/2015/03/backoff.html).

lib/fog/aws/compute.rb
masstamike: Ah yes, sorry about that. I agree completely.

geemus:
Thanks for the clarification, the API limit part was what I wasn't thinking about I suspect. This overall sounds good, just let me know how the testing pans out and then once the puts is fixed I think we should be good to go. Thanks!

geemus:
@masstamike Thanks!

PR #412, opened 2017-11-06
Resolving issue where `Fog::Json` was called instead of `Fog::JSON`.

mgarrick:
`Fog::Json` doesn't exist, so the call to `Fog::Json.encode` failed when providing a `ClientContext` as parameter to `Fog::Lambda#invoke`.

Changing to `Fog::JSON` (all-capitals) resolved this issue.


lanej:
Thanks @mgarrick. Ref: https://github.com/fog/fog-json/blob/03b8cf7/lib/fog/json.rb#L12

geemus:
Thanks!

PR #414, opened 2017-11-15
Avoid creating connection if region is not nil

### LINKED ISSUE #413: connect_write timeout on AWS CodeBuild
I use `fog-aws` via [AssetSync/asset_sync](https://github.com/AssetSync/asset_sync) in `docker build` on AWS CodeBuild.
I set env `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` for build.
But I got a message below.
```
[fog][WARNING] Unable to fetch credentials: connect_write timeout reached
```

On AWS CodeBuild, It's seems that we cannot connect to `INSTANCE_METADATA_HOST`.
I think we need to avoid connecting to `INSTANCE_METADATA_HOST` on it.

hideto0710:
fix #413


geemus:
👍  Thanks!

PR #421, opened 2017-12-08
simple, target and step policies support for asg

starbeast:


geemus:
Overall looks good, though the map keys stuff adds some complexity I can certainly see the appeal.

@lanej could you review also and see what you think? Thanks!

starbeast:
I agree it adds extra complexity but without it we have an inconsistency in terms of how we can define plain model attributes - aws names or snake case symbols, and nested structures which force end users to specify keys with aws names only.

lib/fog/aws/models/auto_scaling/group.rb
```diff
@@ -99,6 +99,10 @@ def enable_metrics_collection(granularity = '1Minute', metrics = {})
           reload
         end
 
+        def policies
+          service.policies.all('AutoScalingGroupName' => id)
```
lanej: This should be 

```ruby
service.policies(filters: { 'AutoScalingGroupName' => id })
```

to prevent eager-loading on access.

Issue #227, opened 2016-02-16
Update Dependencies


plribeiro3000:
As @mdarby mentioned [here](https://github.com/fog/fog/pull/3848#issuecomment-184918533) this gem is relying on a really old version of `fog-core` and i would like to ask to remove the pessimistic dependency if possible. That might give us a lot of headaches and conflicts with other providers if they specify different versions.

So this:

``` ruby
spec.add_dependency 'fog-core',  '~> 1.27'
spec.add_dependency 'fog-json',  '~> 1.0'
spec.add_dependency 'fog-xml',   '~> 0.1'
```

would become something like this:

``` ruby
spec.add_dependency 'fog-core',  '>= 1.27'
spec.add_dependency 'fog-json',  '>= 1.0'
spec.add_dependency 'fog-xml',   '>= 0.1'
```

geemus:
Sounds reasonable. @fog/fog-aws any concerns?

lanej:
would `~>` be more appropriate? Semvar for the fog dependencies ?

plribeiro3000:
@lanej The main problem imo is that you are still locking it (event that on a minor version) and since we have (will) many providers i can see conflicts happening really easy because of that.
Ideally all providers should just list dependencies but not enforce a version but sometimes we might need so just need to be extra careful here.

lanej:
@plribeiro3000 you mean in the main fog repository ?

plribeiro3000:
@lanej Both in the main fog repository (regarding all providers) and in the providers (that are going to require at least fog-core and fog-json or fog-xml).
This 2 level of dependency might turn into a headache if we don't control the big picture.
Like latest `fog-aws` depends on `fog-core ~> 1.27` while lates `fog-whatever` depends on `fog-core >= 1.34`.

lanej:
My only concern would be backwards incompatible changes to the shared fog libraries (fog-core, fog-json, etc.).  In that case wouldn't necessarily want the latest of the providers but I guess that is something you could coordinate.  

If `fog-aws` required `fog-core ~> 1.27` and `fog-softlayer` required `fog-core >= 1.34`, then there shouldn't be a conflict.  As long as another dependency did not declare a need for a `2.x` version.

plribeiro3000:
@lanej True. So if there is backwards incompatible changes you should not pin to a version of `fog-core`  but instead blacklist all the old versions that are not supposed to work against your latest version anymore (like `>= 1.27.0`)

[...]

PR #433, opened 2018-02-26
fog-core 2.x, fog-json 1.x

lanej:
References #227

### DIFF
--- a/lib/fog/aws/models/compute/volume.rb
+++ b/lib/fog/aws/models/compute/volume.rb
@@ -59,25 +59,19 @@ module Fog
             requires :availability_zone
             requires_one :size, :snapshot_id
 
-            if type == 'io1'
-              requires :iops
-            end
+            requires :iops if type == 'io1'
 
             data = service.create_volume(availability_zone, size, create_params).body
             merge_attributes(data)
 
             if tags = self.tags
               # expect eventual consistency
-              Fog.wait_for { self.reload rescue nil }
-              service.create_tags(
-                self.identity,
-                tags
-              )
+              Fog.wait_for { service.volumes.get(identity) }
+              service.create_tags(identity, tags)
             end
 
-            if @server
-              self.server = @server
-            end
+            attach(@service, device) if @server
+
             true
           end
         end
@@ -87,13 +81,7 @@ module Fog
           service.servers.get(server_id)
         end
 
-        def server=(new_server)
-          if new_server
-            attach(new_server)
-          else
-            detach
-          end
-        end
+        attr_writer :server
 
         def snapshots
           requires :id
@@ -109,22 +97,15 @@ module Fog
           detach(true)
         end
 
-        private
-
-        def attachmentSet=(new_attachment_set)
-          merge_attributes(new_attachment_set.first || {})
-        end
-
-        def attach(new_server)
+        def attach(new_server, new_device)
           if !persisted?
             @server = new_server
             self.availability_zone = new_server.availability_zone
           elsif new_server
-            requires :device
             wait_for { ready? }
             @server = nil
             self.server_id = new_server.id
-            service.attach_volume(server_id, id, device)
+            service.attach_volume(server_id, id, new_device)
             reload
           end
         end
@@ -136,6 +117,12 @@ module Fog
             service.detach_volume(id, 'Force' => force)
             reload
           end
+        end
+
+        private
+
+        def attachmentSet=(new_attachment_set)
+          merge_attributes(new_attachment_set.first || {})
         end
 
         def create_params
--- a/lib/fog/aws/models/elb/load_balancer.rb
+++ b/lib/fog/aws/models/elb/load_balancer.rb
@@ -223,15 +223,23 @@ module Fog
         end
 
         def reload
-          super
           @instance_health = nil
           @policy_descriptions = nil
-          self
+          super
         end
 
         def destroy
           requires :id
           service.delete_load_balancer(id)
+        end
+
+        protected
+
+        def all_associations_and_attributes
+          super.merge(
+            'ListenerDescriptions' => attributes['ListenerDescriptions'],
+            'BackendServerDescriptions' => attributes['BackendServerDescriptions'],
+          )
         end
       end
     end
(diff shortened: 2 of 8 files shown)


geemus:
Looks good in theory, seems there are a couple breakages though, unfortunately.

lib/fog/aws/models/elb/load_balancer.rb
```diff
@@ -223,16 +223,24 @@ def save
...
         end
+
+        protected
+
+        def all_associations_and_attributes
+          super.merge(
+            'ListenerDescriptions' => attributes['ListenerDescriptions'],
+            'BackendServerDescriptions' => attributes['BackendServerDescriptions'],
+          )
+        end
```
plribeiro3000: Why do you need this? Maybe we need to improve this method on `fog-core`?

lanej: Attributes are not registered in by an `attribute` declaration but are used to collections from the underlying attributes hash.

This seems like a reasonable workaround using normal OO methodology.

plribeiro3000: Got it. I'm not really pointing this as a problem but i'm more interested in knowing the reasons behind it. If it does inflect an in improvement in `fog-core` i would be more then happy to do it myself. =)

Are those attributes not being used by the DSL for a particular reason? Is there something i can improve in `fog-core` to help improve the usage?

jaredbeck:
Copying the test failure here for convenience:

```
      tests/models/compute/volume_tests.rb
      MissingParameter => The request must contain the parameter device (Fog::Compute::AWS::Error)
        /home/travis/build/fog/fog-aws/lib/fog/aws/requests/compute/attach_volume.rb:77:in `attach_volume'
        /home/travis/build/fog/fog-aws/lib/fog/aws/models/compute/volume.rb:127:in `attach'
        /home/travis/build/fog/fog-aws/lib/fog/aws/models/compute/volume.rb:92:in `server='
        tests/models/compute/volume_tests.rb:11:in `block (3 levels) in <top (required)>'
...
```

Question: Is this test failure the result of breaking changes in the updated dependencies (eg. fog-core) or is it a result of the preceding warning:

```
  AWS | credentials (aws) +++++[fog][WARNING] Unable to fetch credentials: Expected(200) <=> Actual(404 Not Found)
```

I'm not seeing any mention of "volumes" in the fog-core [changelog](https://github.com/fog/fog-core/blob/master/changelog.md) There were actually very few changes between 1.45 and 2.0

```
564ea20 geemus     2018-01-03 09:23:05 -0600  (tag: v2.0.0) v2.0.0
f5521c2 Paul..eiro 2017-11-13 15:55:42 -0200  Merge pull request #221 from NARKOZ/mime-types
94f76f7 Niha..asov 2017-11-13 16:27:30 +0400  Don't check mime/types for load error
a32e960 Niha..asov 2017-11-13 12:49:55 +0400  Add missing runtime dependency declaration to gemspec
65e33d0 Paul..eiro 2017-10-27 11:14:16 -0200  Merge pull request #220 from JustMcAfee/master
e7e0657 JustMcAfee 2017-10-27 08:30:41 -0400  Updating net-ssh parameter names
9fabe1c Wesl..eary 2017-10-26 08:36:45 -0500  Merge pull request #219 from tbrisker/ruby2
5c53ee1 Tome..sker 2017-10-24 11:48:27 +0300  Drop Ruby<2 support
1af65f0 Paul..eiro 2017-10-02 17:54:56 -0300  Merge pull request #218 from fog/fix-reload-for-associations
85cfeae Paul..eiro 2017-09-30 15:12:53 -0300  fix association reload
13979be geemus     2017-08-01 11:46:25 -0500  (tag: v1.45.0) v1.45.0
```

geemus:
I don't think fog-core has a notion of volumes at all, so I don't think it would be that. Other changes to fog-core could be causing this, I suppose, but I'm not sure which one specifically. I suppose the association reload fix might relate, but I'm not at all certain about that (none of the others look particularly related).

geemus:
@jaredbeck no worries. I think it is probably unrelated (and it's just a warning). Most likely there is an actual issue in the parameters passed within that attach volumes call.

lanej:
@geemus @plribeiro3000 

got it working.  needed to change the volume interface slightly, but I think it's for the better.

lib/fog/aws/models/elb/load_balancer.rb
plribeiro3000: Hmmm. So you mean an `attribute` that ideally does not belong to this resource but the API expects? Like a concept issue from the API standpoint?

BTW, this is not a blocker. I'm just trying to figure out the use case to see if/how we can improve `fog-core`. My main goal now is to get it to a better state.

plribeiro3000:
LGTM!

lib/fog/aws/models/compute/volume.rb
```diff
@@ -109,22 +97,15 @@ def force_detach
...
         end
 
-        private
-
-        def attachmentSet=(new_attachment_set)
-          merge_attributes(new_attachment_set.first || {})
-        end
-
-        def attach(new_server)
+        def attach(new_server, new_device)
```
geemus: Could you elaborate on why the new argument is needed? The rest looks good, but I don't think I'm quite groking this. Thanks!

lanej: Tests asserted this procedure

* `v = service.volumes.create(device: '/dev/sda1', **)`
* `v.server = service.instances.create(**) # requires :device`

In step 1, a device is provided to a volume instance. In step 2, the model calls `Fog.wait_for { ready? }` (implicit reload) and then attaches the server.

The issue is between step 1 and 2, there is a property of the local model (`device = '/dev/sda1'`) that does not match the remote model (`device = nil`).  

On `reload`: 

* `fog-core 1.x`, the `device` attribute retained it's value.  
* `fog-core 2.x`, `device` is set the correct remote value (`nil`).

IMO a `reload` should put the local model in the same state as the remote model.  To enforce that, I've removed the ability for `volume.server = server` to attach volume and instead require calling `volume.attach(server, device)`.

jaredbeck:
Thanks @lanej et al. for your work on this.

PR #434, opened 2018-03-14
Include link to full documentation

kylefox:
Unless I've completely missed it, I can't easily find a link to the (excellent) documentation for `fog-aws` — it'd be nice to include it directly in the README.


lanej:
Good catch @kylefox

geemus:
Thanks!

PR #436, opened 2018-03-22
SetInstanceProtection endpoint for auto scaling groups support

thorn:
This PR adds support for SetInstanceProtection endpoint.

https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-instance-termination.html#instance-protection
https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_SetInstanceProtection.html


geemus:
LGTM. @lanej - mind taking an extra peek? Thanks!

PR #438, opened 2018-03-27
Add encryption to EFS FileSystem creation

acant:
These parameters are:
* encrypted
* kms_key_id

Both are optional, but encrypted MUST be true if kms_key_id is specified.

I did not add the additional fields to the tests because the PerformanceMode was not included in the tests, and I  was trying to keep things consistent. But I am happy to update the tests, and I can try to add overriding the PreformanceMode field as well.

### DIFF
--- a/lib/fog/aws/models/efs/file_system.rb
+++ b/lib/fog/aws/models/efs/file_system.rb
@@ -7,6 +7,8 @@ module Fog
         attribute :owner_id,                :aliases => 'OwnerId'
         attribute :creation_token,          :aliases => 'CreationToken'
         attribute :performance_mode,        :aliases => 'PerformanceMode'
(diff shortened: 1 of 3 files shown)


geemus:
Looks reasonable to me, so far anyway. Not certain what should be done around testing.

@lanej could you review as well and advise on testing? Thanks!

acant:
I rebased and squashed by commits, so that they are ready to merge once they are green.

acant:
And something failed. I will look into this tomorrow and fix whatever I have done wrong.

acant:
@geemus sorry for the delay, but I had a hard time getting the test suite running locally.
I updated the tests/requests/efs/helper.rb to use Nullable values, and now it seems to work as expected.

geemus:
@acant no worries, looks good, thanks!

PR #445, opened 2018-04-28
Support new Paris and AP Osaka load balancers in DNS

mattheworiordan:
Sadly these need to be manually updated.
Fortunately it's now simpler as Amazon publish this info.


geemus:
Nice, I'm glad there is at least an easier reference place, though I also really wish you could fetch this programmatically...

PR #446, opened 2018-05-29
Retry if  instance not found when adding EC2 tags

tracemeyers:
Retry is necessary because AWS is eventually consistent. Not doing so
occassionally results in instance tags not being applied.  The specific
error is:
  The instance ID 'i-0e800893aa8d116a1' does not exist (Fog::Compute::AWS::NotFound)

Refer to AWS docs regarding eventual consistency's
InvalidInstanceID.NotFound error

### DIFF
--- a/lib/fog/aws/models/compute/servers.rb
+++ b/lib/fog/aws/models/compute/servers.rb
@@ -162,6 +162,7 @@ module Fog
             server.merge_attributes(instance_set)
             # expect eventual consistency
             if (tags = server.tags) && tags.size > 0
+              Fog.wait_for { server.reload rescue nil }
               Fog.wait_for {
                 begin
                   service.create_tags(server.identity, tags)
--- a/lib/fog/aws/models/compute/servers.rb
+++ b/lib/fog/aws/models/compute/servers.rb
@@ -162,7 +162,6 @@ module Fog
             server.merge_attributes(instance_set)
             # expect eventual consistency
             if (tags = server.tags) && tags.size > 0
-              Fog.wait_for { server.reload rescue nil }
               Fog.wait_for {
                 begin
                   service.create_tags(server.identity, tags)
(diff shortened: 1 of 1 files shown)


lib/fog/aws/models/compute/servers.rb
```diff
@@ -163,10 +163,13 @@ def save_many(template, min_servers = 1, max_servers = nil)
             # expect eventual consistency
             if (tags = server.tags) && tags.size > 0
               Fog.wait_for { server.reload rescue nil }
```
geemus: I believe the original code was trying to use the wait_for on reload in the hopes that it would then be available for the create_tags. As you point out, this is insufficient, so we really want to wait_for the tag create.

That said, I think we could just drop this and only wait_for the create_tag now, what do you think?

tracemeyers: Sounds reasonable to me.  I didn't know if server.reload did something more so if you think it isn't needed then I can try it without it.

geemus: Yeah, my read of it is that it was to protect the create_tags, not for some other side effect. But I could certainly be wrong, seems like it could be worth a shot to simplify and streamline though. Thanks!

geemus:
Thanks, looks like there may be failing tests though, could you look at those as well? Thanks!

tracemeyers:
Sorry, dumb mistake not running the tests. I tried it manually and it worked in my scenario.

Would you mind helping to guide a bit? I've been looking at this but I don't understand enough to know if this is a problem with the test setup or an actual problem (I don't know Ruby). I looked at the reload method and I don't see anything that helps me understand what's going on.

geemus:
I reviewed things a bit more and I think I have an easy fix at least. Your first commit did pass, and it was only after we removed that reload that it began to fail. I suspect this may be a behavior of the mocked version of things that might not appear in real usage. That said, I don't think having the reload there should hurt anything (it just might be redundant). So, how about you just pull out the second commit, putting the old reload back? I believe that should get us back to passing tests and I should be able to pull this in. Does that make sense? Thanks!

tracemeyers:
Thanks for the review. Yes, that makes sense so I'll go ahead with putting the reload back in.

geemus:
Sounds good, as long as that goes back to passing tests (as we expect), I'll try to get it merged asap.

geemus:
Thanks!

PR #449, opened 2018-06-20
Fix c1.xlarge cpu count

romaintb:
Hey, just found a little error, c1.xlarge flavors actually have 8 cpus, not 20.

See : https://aws.amazon.com/ec2/previous-generation/
screen shot 2018-06-20 at 15 30 17


romaintb:
Thanks for the quick merge :)

PR #450, opened 2018-06-20
Update aws flavors cpu count and ram capacity.

xward:
Updated from data available here: https://aws.amazon.com/ec2/pricing/on-demand/ and https://aws.amazon.com/ec2/previous-generation/

I hope I didn't made any mistake :)


geemus:
@xward just wondered if there was anything you needed help with here, as you seem to have created and closed. Thanks!

xward:
@geemus I want to split it better in several pool request, I found that ~ 30 flavor are out of date, and there is ~ 30 missing. I'll pushed the PR soon.

Though, I have one question: ram value is in Megabyte ?

Thank !

geemus:
Yeah, I think I used megabyte values so that they would be integers (instead of needing to have decimal values with gigabytes). So that sounds probably right, though it was a long time ago, so I might not remember accurately.

Issue #448, opened 2018-06-19
Parameterize "RequestLimitExceeded" jitter magnitude


lvangool:
We have a use case where we need greater diversity of jitter (in the case of RequestLimitExceeded). 
The use case is essentially management of multiple AWS resources on behalf of many customers / we're seeing that the jitter differential doesn't handle the burst requests (taking a long time to resolve)

It is currently configured as:
```
jitter = rand(100)
wait_time = ((2.0 ** (1.0 + retries) * 100) + jitter) / 1000.0
```
I'd like to configure something like (pseudo)
```
jitter = rand(100)
jitter_magnitude = @config[:jitter_magnitude] || 1
jitter_adjusted = (jitter / 1000.0) * jitter_magnitude
wait_time = ((2.0 ** (1.0 + retries) * 100) / 1000.0) + jitter_adjusted
```
Any objections?

lvangool:
Add PR: https://github.com/fog/fog-aws/pull/451 
Thanks!

geemus:
closed by #451

PR #451, opened 2018-06-20
Parameterized retry

lvangool:
To address: https://github.com/fog/fog-aws/issues/448
#1 `retry_if_over_limit` to disable retries fully
#2 `retry_jitter_magnitude` to provide a magnitude to jitter (default same)
#3 replace `splin lock` with `sleep`

### DIFF
--- a/lib/fog/aws/compute.rb
+++ b/lib/fog/aws/compute.rb
@@ -6,7 +6,7 @@ module Fog
       class RequestLimitExceeded < Fog::Errors::Error; end
 
       requires :aws_access_key_id, :aws_secret_access_key
-      recognizes :endpoint, :region, :host, :path, :port, :scheme, :persistent, :aws_session_token, :use_iam_profile, :aws_credentials_expire_at, :instrumentor, :instrumentor_name, :version
+      recognizes :endpoint, :region, :host, :path, :port, :scheme, :persistent, :aws_session_token, :use_iam_profile, :aws_credentials_expire_at, :instrumentor, :instrumentor_name, :version, :retry_if_over_limit, :retry_jitter_magnitude
 
       secrets    :aws_secret_access_key, :hmac, :aws_session_token
 
@@ -551,6 +551,8 @@ module Fog
           @instrumentor           = options[:instrumentor]
           @instrumentor_name      = options[:instrumentor_name] || 'fog.aws.compute'
           @version                = options[:version]     ||  '2016-11-15'
+          @retry_if_over_limit    = options.has_key?(:retry_if_over_limit) ? options[:retry_if_over_limit] : true
+          @retry_jitter_magnitude = options[:retry_jitter_magnitude] || 0.1
 
           @use_iam_profile = options[:use_iam_profile]
           setup_credentials(options)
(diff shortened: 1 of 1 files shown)


lib/fog/aws/compute.rb
```diff
@@ -551,6 +551,8 @@ def initialize(options={})
           @instrumentor           = options[:instrumentor]
           @instrumentor_name      = options[:instrumentor_name] || 'fog.aws.compute'
           @version                = options[:version]     ||  '2016-11-15'
+          @retry_if_over_limit    = options.has_key?(:retry_if_over_limit) ? options[:retry_if_over_limit] : true
```
geemus: I wonder if a clearer name might be `retry_request_limit_exceeded`? What do you think?

lib/fog/aws/compute.rb
```diff
@@ -633,18 +635,12 @@ def _request(body, headers, idempotent, parser, retries = 0)
...
                   Fog::Compute::AWS::NotFound.slurp(error, match[:message])
-                when 'RequestLimitExceeded'
-                  if retries < max_retries
-                    jitter = rand(100)
-                    waiting = true
-                    start_time = Time.now
-                    wait_time = ((2.0 ** (1.0 + retries) * 100) + jitter) / 1000.0
+                when 'RequestLimitExceeded'                  
+                  if @retry_if_over_limit && retries < max_retries
+                    jitter = rand * 10 * @retry_jitter_magnitude
```
geemus: The reworking of this math seems to not want to work out in my brain, so I just want to double check I'm interpreting correctly.

I believe in both cases, it should end up that jitter adds some amount between 0 and 1 second to the resulting value (with the default magnitude). The before/after just change where the multiplication/divisions happen to reach this result. Does that match with your understanding/intention?

lib/fog/aws/compute.rb
lvangool: Yup, happy with `retry_request_limit_exceeded` - got hit by: https://martinfowler.com/bliki/TwoHardThings.html

lib/fog/aws/compute.rb
lvangool: You're right - 
Original code - the jitter diff 0.01 - 0.1 secs (midpoint 0.05 secs)
An new code - the jitter diff 0.1 - 1 secs (midpoint 0.5 secs) (default)

This makes a bit more sense IMHO given AWS throttling happens at second granularity. 
But happy to change the default to maintain consistency if you prefer?

Ps. I tried to lay it out in such a way that the magnitude makes sense. ie. 
magnitute 0.01 --> jitter midpoint 0.05 secs
magnitute 0.1 (default) --> jitter midpoint 0.5 secs
magnitute 1 --> jitter midpoint 5 secs
magnitute 10 (default) --> jitter midpoint 50 secs
etc.

geemus: Thanks for laying that out, I thought there might be a difference in values, but my brain was really not cooperating in the specifics. That seems fine, thanks for the detailed explanation.

lvangool:
Thanks :)

lvangool:
@geemus any plans to cut a new gem version with this? :)

lib/fog/aws/compute.rb
```diff
@@ -546,11 +546,13 @@ class Real
...
-          @region                 = options[:region] ||= 'us-east-1'
-          @instrumentor           = options[:instrumentor]
-          @instrumentor_name      = options[:instrumentor_name] || 'fog.aws.compute'
-          @version                = options[:version]     ||  '2016-11-15'
+          @connection_options           = options[:connection_options] || {}
+          @region                       = options[:region] ||= 'us-east-1'
+          @instrumentor                 = options[:instrumentor]
+          @instrumentor_name            = options[:instrumentor_name] || 'fog.aws.compute'
+          @version                      = options[:version]     ||  '2016-11-15'
+          @retry_request_limit_exceeded = options.has_key?(:retry_request_limit_exceeded) ? options[:retry_request_limit_exceeded] : true
```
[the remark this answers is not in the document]
lvangool: 👍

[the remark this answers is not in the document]
lvangool: PR for that here: https://github.com/fog/fog-aws/pull/457

geemus:
@lvangool releases are, more often than not, just when somebody asks nicely. So we can certainly try to get that done soon.

@lanej would you be up for cutting the release? My only concern would be that we had some help on improving the changelog, but I think the release process might have scripting in place that would clobber those manual changes? Anyway, would appreciate whatever insights you might have about that, even if you need me to cut the actual release.

lvangool:
Thanks @geemus much appreciated.

PR #452, opened 2018-06-21
Update aws flavors cpu count.

xward:
Updated from data available here:
- https://aws.amazon.com/ec2/pricing/on-demand/
- https://aws.amazon.com/ec2/previous-generation/

Cheers !


xward:
Actually the question is : what is the unit of core ? ecu ou vcpu ?

geemus:
Good question. I don't recall for sure (and it seems strange that we would have so many that are off). The old values don't really obviously match with either ecu or vcpu either. So, I'm not sure what it used to be. I think standardizing to vcpu makes sense (especially since so many just have "variable" for ecu). So I'm fine with the change as you've done, thanks!

PR #457, opened 2018-07-04
Simplify constructor

lvangool:
As suggested here: https://github.com/fog/fog-aws/pull/451
```
options.has_key?(:retry_request_limit_exceeded) ? options[:retry_request_limit_exceeded] : true
```
becomes:
```
options.fetch(:retry_request_limit_exceeded, true)
```


geemus:
Thanks!

PR #460, opened 2018-08-02
Fix t1.micro bits

tas50:
It's not a 0 bit instance. It's a 32bit instance.

Signed-off-by: Tim Smith <tsmith@chef.io>


geemus:
Thanks!

PR #463, opened 2018-08-24
Escape / in presigned URLs

### LINKED ISSUE #462: s3: fog returns bad URL (with correct signature)
### Context
 * Ceph RadosGW `12.2.7`
 * fog `3.0.0`

### Background
We're using fog with a non-aws S3 implementation (namely the Ceph RadosGW) and it turns out that fog returns non-working URLs when `aws_signature_version` set to 4.

[AWS' Signature v4 documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) states that:

> The X-Amz-Credential value in the URL shows the "/" character only for readability. In practice, it should be encoded as %2F. For example:

Despites having `fog` generate a valid signature (with escaped slashes in the query string), the string returned when invoking `connection.directories[0].files[0].url(Time.now + 1000)` isn't escaped.

### Recipe to reproduce
* Against a Ceph RadosGW implementation (ceph-nano for instance https://github.com/ceph/cn) run the following snippet:

```ruby
require 'fog-aws'
connection = Fog::Storage.new({
  :provider                 => 'AWS',
  :aws_access_key_id        => 'xxxxxxx',
  :aws_secret_access_key    => 'yyyyyyyy', 
  :endpoint => 'http://127.0.0.1:9000',
  :path_style => true,
  :aws_signature_version => 4
  })
connection.directories[0].files[0].url(Time.now + 1000)
```

* Open the URL generated by fog, you will get the following output:

```xml
<Error>
<Code>SignatureDoesNotMatch</Code>
<RequestId>tx0000000000000000bb469-005b802afd-1b01774-default</RequestId>
<HostId>1b01774-default-default</HostId>
</Error>
```

* Replace `/` by `%2F` in the generated URL, you should be able to read the object.

### Source of the issue
 * The v4 signature is computed properly 
 * When the generated URL is returned, it is not escaped, which causes the aforementioned error

alexcern:
As per https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

Fixes #462

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -316,7 +316,8 @@ module Fog
         def params_to_url(params)
           query = params[:query] && params[:query].map do |key, value|
             if value
-              [key, escape(value.to_s)].join('=')
+              # URL parameters need / to be escaped
+              [key, Fog::AWS.escape(value.to_s)].join('=')
             else
               key
             end


lanej:
```
The X-Amz-Credential value in the URL shows the "/" character only for readability. In practice, it should be encoded as %2F:
```

Makes sense to me.  @geemus?

geemus:
@lanej Sounds good to me. Thanks for checking in.

@alexcern thanks for the detailed report and fix!

Issue #466, opened 2018-09-05
How to fix deprecation warning: "The format Fog::CDN::AWS is deprecated"


jaredbeck:
Hi, I'd like to help fix this warning:

```
be rake -T > /dev/null
[fog][DEPRECATION] Unable to load Fog::AWS::CDN
[fog][DEPRECATION] The format Fog::CDN::AWS is deprecated
[fog][DEPRECATION] Unable to load Fog::AWS::Compute
[fog][DEPRECATION] The format Fog::Compute::AWS is deprecated
[fog][DEPRECATION] Unable to load Fog::AWS::DNS
[fog][DEPRECATION] The format Fog::DNS::AWS is deprecated
[fog][DEPRECATION] Unable to load Fog::AWS::Storage
[fog][DEPRECATION] The format Fog::Storage::AWS is deprecated
```

Do we need to eg. `mv lib/fog lib/aws`? How can I help?

See also https://github.com/fog/fog-core/pull/241

```bash
grep fog Gemfile
gem 'fog-aws'

bundle | grep fog
Using fog-core 2.1.2
Using fog-json 1.2.0
Using fog-xml 0.1.3
Using fog-aws 3.0.0
```

Jabernathy88:
Me too.

plribeiro3000:
You will need to change the convention from `Fog::CDN::AWS` to `Fog::AWS::CDN`.

The issue is that there is no convention among the providers so we have both ways defined here and there.

So we are deprecating the first one in favor of the second one. That should allow us to have the same implementation in all providers.

jaredbeck:
> You will need to change the convention from Fog::CDN::AWS to Fog::AWS::CDN.

Is this something you think I can tackle? If so, I'll attempt a PR. Do you want to try to define both, or are you willing to make a breaking change (ie. release new major version)?

plribeiro3000:
It will have to be a breaking change so a new major version will be
released.

I believe you can tackle it. It just a change in the name. =)


Em qua, 5 de set de 2018 19:27, Jared Beck <notifications@github.com>
escreveu:

geemus:
It might be fine to have both? I think the issue at hand here is that, by default, we require the deprecated version. I think we could probably do this?

1. Change to use the new name for the class.
2. Change existing requires to use this new name.
3. Add a new class that just aliases (and will raise the warning if anybody includes directly).

I think that way we could perhaps avoid the requirement of a major version bump.

@plribeiro3000 what do you think?

[...]

PR #467, opened 2018-09-06
Rename CDN::AWS to AWS::CDN

jaredbeck:
Fixes fog-core deprecation warning

```
[fog][DEPRECATION] Unable to load Fog::AWS::CDN
[fog][DEPRECATION] The format Fog::CDN::AWS is deprecated
```

Part of https://github.com/fog/fog-aws/issues/466

### DIFF
--- a/lib/fog/aws.rb
+++ b/lib/fog/aws.rb
@@ -5,10 +5,6 @@ require 'fog/json'
 require File.expand_path('../aws/version', __FILE__)
 
 module Fog
-  module CDN
-    autoload :AWS,  File.expand_path('../aws/cdn', __FILE__)
-  end
-
   module Compute
     autoload :AWS, File.expand_path('../aws/compute', __FILE__)
   end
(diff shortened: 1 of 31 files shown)


plribeiro3000:
👍 from me.

Need someone active on this repo to look and merge tho

jaredbeck:
How should I run tests locally? I couldn't find any docs on it. So, I just pushed and hoped for CI to pass.

jaredbeck:
Also,

> Add a new class that just aliases (and will raise the warning if anybody includes directly).

I don't know how to do a warning. ActiveSupport has `deprecate_constant` but I assumed you don't want that dependency.

plribeiro3000:
I believe you can just overwrite the constructor raising a deprecation notice and call super in the next line.

plribeiro3000:
Yeap, the construction you made is not gonna work.

You will have to open the class and inherit manually then overwrite the method initialize.

If you take a look at the `Fog::Service` module you will see it has a `new` method being defined [here](https://github.com/fog/fog-core/blob/master/lib/fog/core/service.rb#L52-L117)

Sou you can open the class and define it like:

```ruby
class Fog::CDN::AWS < Fog::AWS::CDN
  def self.new(config = {})
    Fog::Logger.deprecation 'Your message here'
    super(config)
  end
end
```

jaredbeck:
> You will have to open the class and inherit manually then overwrite the method initialize.

Done. Thanks.

geemus:
Looks good to me, @plribeiro3000 does that match with what you were asking?

PR #469, opened 2018-09-17
Rename DNS::AWS to AWS::DNS

sue445:
Fixes fog-core deprecation warning

```
[fog][DEPRECATION] Unable to load Fog::AWS::DNS
[fog][DEPRECATION] The format Fog::DNS::AWS is deprecated
```

Part of #466

### DIFF
--- a/lib/fog/aws.rb
+++ b/lib/fog/aws.rb
@@ -13,10 +13,6 @@ module Fog
     autoload :AWS, File.expand_path('../aws/compute', __FILE__)
   end
 
-  module DNS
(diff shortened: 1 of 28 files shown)


geemus:
@sue445 looks good, just noticed that it looks like merging your first changeset must have introduced a conflict with this one. Could you help resolve that so that I can bring this in?

geemus:
@sue445 never mind, pretty sure I got it. As long as tests pass again, I think we should be good to go. Thanks!

geemus:
Thanks!

PR #481, opened 2018-10-05
Updated EC2 flavors to include all instance types

gonoph:
Using the json data provided by the open source project
[ec2instances][1], I was able to programmatically pull in all the
current instance types with their metadata.

[1]: https://github.com/powdahound/ec2instances.info

Also included is the ruby generator script and the .erb template of the flavors.rb file.

Be kind, it's my first ruby script.

Ideally, the generator script would be called to generate the flavors.rb file prior to the packaging and testing. I'm not familiar with travis and ruby in order to make that happen.

### DIFF
--- a/scripts/flavors.erb
+++ b/scripts/flavors.erb
@@ -1,0 +1,235 @@
+require 'fog/aws/models/compute/flavor'
+
+module Fog
+  module AWS
+    class Compute
+      FLAVORS = [
+<% instances.each do |item| %>        {
+          :id                      => '<%= item[:id] %>',
+          :name                    => '<%= item[:name] %>',
+          :bits                    => <%= item[:bits] %>,
+          :cores                   => <%= item[:cores] %>,
+          :disk                    => <%= item[:disk] %>,
+          :ram                     => <%= item[:ram] %>,
+          :ebs_optimized_available => <%= item[:ebs_optimized_available] %>,
+          :instance_store_volumes  => <%= item[:instance_store_volumes] %>
+        }<% unless item.equal?(instances.last) %>,<% end %>
+<% end %>      ]
+
+      class Flavors < Fog::Collection
+        model Fog::AWS::Compute::Flavor
+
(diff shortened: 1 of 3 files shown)


scripts/flavors.erb
```diff
@@ -0,0 +1,235 @@
+require 'fog/aws/models/compute/flavor'
+
+module Fog
```
geemus: It would probably be ok to leave this file out. I don't think we really need to build objects in order to do the formatting stuff. I think just filling in the templates directly from the data would probably make it easier to maintain/debug as needed. Does that make sense?

scripts/make_flavors.rb
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/ruby
```
geemus: Looks like a great start. Rather than `/scripts/make_flavors`, it would probably make sense to add this as a rake task instead. I think that is more in fitting with normal Ruby paradigms anyway. Does that make sense?

scripts/flavors.erb
gonoph: Let me rephrase what I think you're trying to say, to make sure I understand. You're saying to not include the .erb template, but to write out a static file directly from the sorted hashed set?

In my mind, I was trying to introduce as little change as possible, probably at some point including a build step that generates the model file before the actual tests and packaging.

scripts/make_flavors.rb
gonoph: That does make sense, but I'll need to read up on the "proper way" of doing this. First attempt at ruby.

Do you see the rake task being ran as part of the build process, or as part of the installation process on the host?

geemus: No worries, feel free to ask questions as you have them. I appreciate you working through this with me. I would imagine it would happen on the development side and/or as part of the build process, rather than during installation. So the release would have the output bundled with it, to keep things relatively simple (though it would require a new release to add now info, this is the state we are already in).

PR #488, opened 2018-10-24
Add ability to force delete a bucket with objects

ramonpm:


ramonpm:
Yeah Geemus, it makes sense, it works for my case where the amount is not so big. Will improve it later, then I reopen this.

ramonpm:
Actually, take a closer look at the `#destroy!` method, it will retry after some time, if the scenario involves a very big amount of objects then it would make sense to give a higher number of max attempts and initial wait time, it should work.

ramonpm:
Manual tests of the `#destroy!` method with the `#clear!` commented.
```
ramonpm@ramonpm ~/workspace/ramonpm/fog-aws $ pry
[1] pry(main)> require './lib/fog-aws.rb'
=> true
[2] pry(main)> name = "test-rpm-mar"
=> "test-rpm-mar"
[3] pry(main)> s3 = Fog::Storage.new(:provider => 'AWS')
[4] pry(main)> dir = s3.directories.get(name)
=>   <Fog::AWS::Storage::Directory
    key="test-rpm-mar",
    creation_date=nil,
    location="us-east-1"
  >
[5] pry(main)> dir.destroy!
Raised BucketNotEmpty. Will try again ...
Raised BucketNotEmpty. Will try again ...
Raised BucketNotEmpty. Will try again ...
Excon::Error::Conflict: Expected(204) <=> Actual(409 Conflict)
excon.error.response
  :body          => "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>BucketNotEmpty</Code><Message>The bucket you tried to delete is not empty</Message><BucketName>test-rpm-mar</BucketName><RequestId>990340CB2F432130</RequestId><HostId>XVSczZ6D3uEeKxB4Rm1UMus9BTkUFWKcZJa0LHwVSHLCTBPAhnhnFm2t87C3USHktXcKkytYkI0=</HostId></Error>"
  :cookies       => [
  ]
  :headers       => {
    "Content-Type"     => "application/xml"
    "Date"             => "Fri, 26 Oct 2018 12:05:22 GMT"
    "Server"           => "AmazonS3"
    "x-amz-id-2"       => "XVSczZ6D3uEeKxB4Rm1UMus9BTkUFWKcZJa0LHwVSHLCTBPAhnhnFm2t87C3USHktXcKkytYkI0="
    "x-amz-request-id" => "990340CB2F432130"
  }
  :host          => "test-rpm-mar.s3.amazonaws.com"
  :local_address => "192.168.25.6"
  :local_port    => 40154
  :path          => "/"
  :port          => 443
  :reason_phrase => "Conflict"
  :remote_ip     => "52.216.17.224"
  :status        => 409
  :status_line   => "HTTP/1.1 409 Conflict\r\n"
from /home/ramonpm/.rvm/gems/ruby-2.3.4/gems/excon-0.62.0/lib/excon/middlewares/expects.rb:7:in `response_call'
[6] pry(main)> 
```

geemus:
It does seem like an improvement. I guess my point would just be that we should be able to ascertain if there are more objects and proactively address them, rather than just having to rely on retries. I think that would result in fewer surprises. What do you think?

ramonpm:
I think that would be good, however, we have an async problem, so if I call clear then check for objects again I can't be sure that the fired delete object requests are finished, can I? I think that's the reason AWS SDK relied on retries as well, that code is pretty copied and pastes with a few adaptations.

geemus:
First off, thanks for working on this and taking time to discuss with me, I appreciate it. I had (once again) forgotten about the possibility of async/eventual-consistency related problems. You are correct that we would be hard pressed to do this and have it work consistently in the way that I stated. Falling back to retries is probably more reliable (even though in really-long-lists it may still fall short). I'll give the particulars another look, but you've convinced me in terms of the overall approach. Thanks!

lib/fog/aws/models/storage/directory.rb
```diff
@@ -22,6 +22,13 @@ def acl=(new_acl)
           end
         end
 
+        def clear!
```
geemus: Do we intend for this to only really be used inside of `destroy!`? I think that probably makes the most sense, and if so maybe we should move it down into the private methods to make that clearer. What do you think?

lib/fog/aws/models/storage/directory.rb
```diff
@@ -30,6 +37,38 @@ def destroy
...
         end
 
+        # @param options [Hash] (defaults to: {}) — a customizable set of options
+        # @option options max_attempts [Integer] — default: 3 — Maximum number of
+        #   times to attempt to delete the empty bucket.
+        # @option options initial_wait [Float] — default: 1.3 — Seconds to wait before
+        #   retrying the call to delete the bucket, exponentially increased for each attempt.
+        def destroy!(options = {})
+          requires :key
+          options = {
```
geemus: I think for consistency it might make sense to convert this to use `Fog.wait_for { ... }`? It has defaults for intervals/back-off already built in and is what other things use. It would change from a number of max_attempts to a timeout though, if I recall correctly. I just thought it might make it easier to understand/adopt/maintain to reuse the existing mechanism if we think it is similar enough. What do you think? The implementation is here: https://github.com/fog/fog-core/blob/7865ef77ea990fd0d085e49c28e15957b7ce0d2b/lib/fog/core/wait_for.rb and spec here: https://github.com/fog/fog-core/blob/7865ef77ea990fd0d085e49c28e15957b7ce0d2b/spec/wait_for_spec.rb

lib/fog/aws/models/storage/directory.rb
ramonpm: Actually, I was thinking that it could be used standalone, that's why I left it public, but yeah, we can move it to private if you think it's better.

lib/fog/aws/models/storage/directory.rb
ramonpm: looks good, will use that instead then

lib/fog/aws/models/storage/directory.rb
geemus: I'm open to it being public, I just thought it might be confusing/surprising since it won't do multiple pages/retries and because people might then not be sure how to differentiate between when to use this and destroy. So yeah, seemed like maybe private was simpler/safer, but I'm certainly open to discussing it. Thanks!

geemus:
Looks good, thanks for working through this with me!

Issue #423, opened 2017-12-15
Accessing AWS S3 using EC2 Instance Profile


suketng:
Hi,

Can you please help me with the code to use to access AWS S3 using the EC2 Instance Profile ? I do not want to use Access Key/Secret Key to access AWS S3.

geemus:
What have you tried so far? What issues are you seeing?

davewalter:
Hi @geemus,

We've been struggling with this same issue recently. So far, we have tried 3 possible configurations.

The first configuration completely omitted the aws_access_key_id and aws_secret_access_key options, and resulted in the following error:

```
ArgumentError: Missing required arguments: aws_access_key_id, aws_secret_access_key
/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/vendor/bundle/ruby/2.4.0/gems/fog-core-1.43.0/lib/fog/core/service.rb:244:in `validate_options'
/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/vendor/bundle/ruby/2.4.0/gems/fog-core-1.43.0/lib/fog/core/service.rb:268:in `handle_settings'
/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/vendor/bundle/ruby/2.4.0/gems/fog-core-1.43.0/lib/fog/core/service.rb:98:in `new'
/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/vendor/bundle/ruby/2.4.0/gems/fog-core-1.43.0/lib/fog/core/services_mixin.rb:16:in `new'
/var/vcap/packages/cloud_controller_ng/cloud_controller_ng/vendor/bundle/ruby/2.4.0/gems/fog-core-1.43.0/lib/fog/storage.rb:27:in `new'
/var/vcap/data/packages/cloud_controller_ng/tmp/cloud_controller_ng/lib/cloud_controller/blobstore/fog/fog_client.rb:194:in `connection'
/var/vcap/data/packages/cloud_controller_ng/tmp/cloud_controller_ng/lib/cloud_controller/blobstore/fog/fog_client.rb:184:in `directory'
/var/vcap/data/packages/cloud_controller_ng/tmp/cloud_controller_ng/lib/cloud_controller/blobstore/fog/fog_client.rb:180:in `dir'
/var/vcap/data/packages/cloud_controller_ng/tmp/cloud_controller_ng/lib/cloud_controller/blobstore/fog/fog_client.rb:130:in `files'
[...]

PR #499, opened 2018-12-11
Do not require access keys in Storage class

surminus:
I wish to use an IAM profile for an instance to grant access to S3 buckets rather than pass in access keys.

This is possible throughout fog-aws by passing the `use_iam_profile`, but the `Storage` class requires aws_access_key_id and aws_secret_access_key parameters.

This causes errors related to malformed headers being submitted to Amazon S3 similar to being described in this issue. [1]

As suggested by the issue, workarounds such as passing in blank or bogus access keys do not resolve the issue.

The error is resolved by avoiding the necessity to pass in the access keys and only pass in the use_iam_profile parameter.

I cannot see an issue with not having an explicit require and I haven't updated any tests, but please let me know if I need to do further work against it.

[1] https://github.com/fog/fog-aws/issues/423#issuecomment-367133454

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -41,8 +41,7 @@ module Fog
         website
       ]
 
-      requires :aws_access_key_id, :aws_secret_access_key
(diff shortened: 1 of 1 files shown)


geemus:
No worries, that line does have some pretty big implications for sure.

What you describe sounds about right to me, and some experimentation sounds like a great way to pin down why it might not be doing what we think. If you have time and ability to do that prying, that would be awesome. Just let me know if you have other questions or things I can help with.

Thanks!

PR #501, opened 2019-01-29
Fix for Aurora Server Provisioning.

lockstone:
1. A non-functional commit, just a comment. I'm not sure the commit from December adding in the check for 'active' in the ready? method of RDS::Cluster is going to work, specifically for Aurora clusters. Please see my comment.
2. Updated the check for the "aurora" engine value in RDS::Server to check if the value starts with "aurora" instead of equaling "aurora." Early versions of Aurora did use just "aurora" for the engine. But that changed when Aurora started supporting more than just MySQL.

### DIFF
--- a/lib/fog/aws/models/rds/cluster.rb
+++ b/lib/fog/aws/models/rds/cluster.rb
@@ -23,7 +23,15 @@ module Fog
         attr_accessor :storage_encrypted #not in the response
 
         def ready?
-          state == "available" || state == 'active'
+          # [2019.01] I don't think this is going to work, at least not with Aurora
+          # clusters. In my testing, the state reported by Fog for an Aurora cluster
+          # is "active" as soon as the cluster is retrievable from AWS, and the
+          # value doesn't change after that. Contrast that with the AWS Console UI,
+          # which reports the cluster as "Creating" while it's being created. I don't
+          # know where Fog is getting the state value from, but I don't think it's
+          # correct, at least not for the purpose of knowing if the Cluster is ready
+          # to have individual instances added to it.
+          state == 'available' || state == 'active'
         end
 
         def snapshots
--- a/lib/fog/aws/models/rds/server.rb
+++ b/lib/fog/aws/models/rds/server.rb
@@ -118,14 +118,14 @@ module Fog
           else
             requires :engine
 
-            if engine == 'aurora'
+            if engine.start_with?('aurora')
               requires :cluster_id
-              self.flavor_id ||= 'db.r3.large'
+              self.flavor_id ||= 'db.r4.large'
             else
               requires :master_username
               requires :password
               requires :allocated_storage
-              self.flavor_id ||= 'db.m1.small'
+              self.flavor_id ||= 'db.m4.large'
             end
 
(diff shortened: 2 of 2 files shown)


lockstone:
The CI failures seem to be related to Bundler. I don't think there's anything I can do about that.

lockstone:
Hi Wesley. Yea, I'm not sure what Fog can do. It's frustrating because it would be great to be able to check the status/state and know whether or not the cluster was ready. AWS's Console UI, whatever it's doing, does know because it shows the cluster in "Creating" state and then "Available" once it's done. So the information is there somewhere, but I don't know enough about how Fog is requesting that info to know if it can do anything.

I changed the default instance types because Amazon has already phased out the m1 line pretty much completely, m3s are mostly gone, too. Even m4s will be on their way out in the not too distant future in favor of m5s. Similarly for r3s, they're pretty much gone in favor of r5s and r4s. Here's a page that details the current generation of RDS instance types.

https://aws.amazon.com/rds/instance-types/

lockstone:
Hmmm, I think we're doing essentially what you describe, though we're not calling "create_db_cluster," at least not directly.

To create the Aurora Cluster, we're using a Fog::AWS::RDS object and calling clusters.create on it. This successfully starts the process of creating the cluster.

We then start calling clusters#get, which, for a few seconds returns nil (which is fine). When it does start returning the Cluster object, the state is "active" right from the start, even though, according to the Console UI, the cluster is still "Creating."

For background, the reason we're trying to get the actual state of the Cluster is that we want to add an RDS instance to the cluster as soon as it's ready to accept an instance. If we try to do it too quickly, while the cluster is still Creating, then it fails. Our workaround is to sleep for 60 seconds after initially requesting the cluster, knowing that 99% of the time that's more than long enough since the cluster generally only takes maybe 15-20 seconds to create.

Is there a better/different way to achieve what we want? Would describe_db_clusters yield a different, more accurate, state value? Or is the clusters.get essentially doing that anyway?

geemus:
Thanks for the further detail, I think we are more or less on the same page, as you suggest. I believe clusters#create calls create_db_cluster underneath, and that clusters#get is calling describe_db_cluster. So it sounds like the issue at hand is in responses, which we have no direct control over. It might be worth escalating this to the RDS team directly, perhaps they know about it or could provide us with other suggestions?

lockstone:
Hi Wesley. This conversation has been mostly about the comment I added to ../aws/models/rds/cluster.rb. Which, obviously, is not a functional change.

The functional change in ../aws/models/rds/server.rb is relevant irrespective of whatever might be happening with the response that Fog is receiving from AWS on the state/status value. Notably, the change from checking for just "aurora" vs. start_with?("aurora") to account for the fact that the engine string AWS now uses to identify Aurora is never just "aurora" anymore.

With that said, can we get this PR merged? Feel free to remove my comment in ../aws/models/rds/cluster.rb if you think that is superfluous or just confuses things. But the change in ../aws/models/rds/server.rb is definitely needed.

geemus:
Yes, sorry. That seems reasonable. Thanks!

lockstone:
Thanks, Wesley!

PR #502, opened 2019-01-29
Add MaxResults filter to describe security groups

KevinLoiseau:
Hello,

In this PR I did:
- Added filters MaxResults and NextToken to describe_security_groups.
- Updated the describe security group parser to add nextToken in the result load.

Thanks in advance for your time


geemus:
Thanks!

PR #503, opened 2019-02-08
Correct DynamoDB update_item method

postmodern:
1. The `AttributeUpdates` option is deprecated and should no longer be passed in by default, unless the deprecated behavior is requested.
2. Updated the documentation to reflect the current [AWS DynamoDB update_item](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/dynamo-example-update-table-item.html) documentation.


geemus:
Looks good, thanks! Just trying to get CI fixed before bringing this in (it looks to be breakage altogether unrelated to your change).

geemus:
Thanks!

postmodern:
:+1:

PR #506, opened 2019-02-21
Fix some requests when S3 acceleration is enabled

NARKOZ:
This PR fixes some methods to make them work with acceleration enabled #505

+ get_bucket_location uses host instead of path_style

+ get_service explicitly uses s3.amazonaws.com as per docs https://docs.aws.amazon.com/AmazonS3/latest/API/RESTServiceGET.html

### DIFF
--- a/lib/fog/aws/requests/storage/get_bucket_location.rb
+++ b/lib/fog/aws/requests/storage/get_bucket_location.rb
@@ -12,7 +12,7 @@ module Fog
         #   * body [Hash]:
         #     * LocationConstraint [String] - Location constraint of the bucket
         #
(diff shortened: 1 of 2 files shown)


lib/fog/aws/requests/storage/get_bucket_location.rb
```diff
@@ -22,8 +22,7 @@ def get_bucket_location(bucket_name)
             :idempotent => true,
             :method   => 'GET',
             :parser   => Fog::Parsers::AWS::Storage::GetBucketLocation.new,
-            :query    => {'location' => nil},
-            :path_style => true
```
geemus: I'm slightly concerned that this will break something, though only because I imagine we set this for a reason at some point in the past (I wasn't readily able to ascertain the history, due to the extraction from fog/fog).

PR #507, opened 2019-03-01
Avoid using bucket_name.host if host is overriden.

deepfryed:
In cases where you have CNAME pointing to the S3 bucket, implicitly combining the bucket name and overriden host prevents signed url generation with CNAME pointing to the bucket.  

This change fixes the issue. Potential gotcha is now you cannot override the hostname for region, but it should allow you to specify the exact hostname (along with bucket name) if needed.

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -43,7 +43,7 @@ module Fog
       ]
 
       requires :aws_access_key_id, :aws_secret_access_key
-      recognizes :endpoint, :region, :host, :port, :scheme, :persistent, :use_iam_profile, :aws_session_token, :aws_credentials_expire_at, :path_style, :acceleration, :instrumentor, :instrumentor_name, :aws_signature_version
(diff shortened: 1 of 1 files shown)


geemus:
@deepfryed It's been a while since I worked on this, so I beg your indulgence if these questions don't make the most sense. Do you think not being able to override hostname could result in a regression for some users? Is there any potential for a 3rd branch of the conditional to still support this use case? Thanks!

deepfryed:
@geemus fair point, I've added separate options to avoid the confusion.

virtual_host: true  => uses the bucket name as virtual host
virtual_host: true, cname: <fqdn> => uses the cname as the virtual host

geemus:
Thanks, I'm not certain it would have actually manifested as a problem, but I tend toward conservative for backwards compatibility.

This should guard us from that, though I think you may need to add `:virtual_host` to the recognizes list here: https://github.com/fog/fog-aws/blob/aacb0a39a162b6a6a1345828a73554e22d369b82/lib/fog/aws/storage.rb#L46

You should see that `:path_style` is already in there. Without adding it, I believe it will throw errors (or at least warnings) that you are passing around unknown things. Does that make sense?

deepfryed:
@geemus thanks, recognised list updated   👍

PR #511, opened 2019-03-25
Enhance/add attribute db subnet group for db instance

KevinLoiseau:
Hello,

In this PR I added the attribute db_subnet_group in both parser and model for RDS servers. The documentation related to this db instance's attribute can be found here: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DBInstance.html


geemus:
Looks good, thanks!

PR #512, opened 2019-03-27
Enhance/handle ELBv2 api version

KevinLoiseau:
Hello,

In this PR I added the possibility to choose the ELB service version and added some attributes on the describe_load_balancer parser corresponding to the newest version.
This changes was implemented to add the compatibility with "ELBv2" which is on the ELB api version ` 2015-12-01`.
- The documentation related to this version is here: https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/Welcome.html
- You can also find the corresponding new LoadBalancer attributes here: https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_LoadBalancer.html 

### DIFF
--- a/lib/fog/aws/elb.rb
+++ b/lib/fog/aws/elb.rb
@@ -14,7 +14,7 @@ module Fog
       class ValidationError             < Fog::Errors::Error; end
 
       requires :aws_access_key_id, :aws_secret_access_key
-      recognizes :region, :host, :path, :port, :scheme, :persistent, :use_iam_profile, :aws_session_token, :aws_credentials_expire_at, :instrumentor, :instrumentor_name
+      recognizes :region, :host, :path, :port, :scheme, :persistent, :use_iam_profile, :aws_session_token, :aws_credentials_expire_at, :version, :instrumentor, :instrumentor_name
 
       request_path 'fog/aws/requests/elb'
       request :configure_health_check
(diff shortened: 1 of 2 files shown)


geemus:
I think it should be safe to simply change the version specified here, rather than needing to have a version option. I believe that's how we have handled it in the past, as versions seem to be backwards compatible and new functionality doesn't come in until things are added (like your parser work). My inclination would be to just change the version here as well, and not have the version option. Does that make sense? What do you think?

KevinLoiseau:
Thank you for have been taking a look to my PR.
Unfortunally this new version is not backwards compatible:
- the version `2012-06-01` is about ELBv1 which are ELB type "classic"
- the version `2015-12-01` is about ELBv2 which are ELB type "application" and "network"
So `describe_load_balancers`  doesn't list same objects according to the selected version.

If you prefere I can create a new service named `ELBv2` which will inherite of `ELB` to be more explicit. What do you think ?

geemus:
Thanks for the clarification, I didn't realize there was such a stark difference for this particular case. I think your approach seems imminently reasonable given that context. The broader question may just be when/if we might want to consider changing the default version from the rather old one to the somewhat newer one (and requiring users to opt-in to the old one if they still want it). I don't see any reason to expand the scope of this particular PR to include that, but I'm open to discussing it further if you would like. Thanks!

Issue #516, opened 2019-05-09
Path style is being deprecated


tedPen:
Wasn't able to find an issue for this but it looks like per Amazon S3's path style will be deprecated effective October 30th, 2020.

Might be worth removing that as an option. I would be willing to do this work if it's something you'd want to pursue.

Edit: Forgot to include source: [Official AWS announcement](https://forums.aws.amazon.com/ann.jspa?annID=6776)

geemus:
Sure, I'm open to it. If I recall there are some characters allowed for bucket names that are not valid in subdomains though? If so, is there some escaping mechanism we need to think about?

tedPen:
Looks like AWS considered that by updating their [naming restrictions](https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html)

>Amazon S3 no longer supports creating bucket names that contain uppercase letters or underscores. This change ensures that each bucket can be addressed using virtual host style addressing

geemus:
Thanks for the deep link, that clarifies a lot. Seems like we could probably start working on removing this to simplify things then.

tedPen:
AWS has backpedaled after feedback from customers: https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/. 

I think it makes sense to remove path_style as an option, but keep in place the existing fallbacks. I created a pull request with change, plan to follow up on tests.

geemus:
@tedPen Sounds good, thanks for the PR and update.

mzch:
I'd like to leave path_style as an option for S3 compatible storage, e.g. Minio.

geemus:
@mzch ah yeah, that is a good call out. Thanks for the reminder.

WilliamDenniss:
+1 to **keep** path_style. It hasn't been removed from AWS after all, and there are a lot of S3-compatible APIs. I need it for Minio too.

geemus:
Closing this as we have good reasons to leave this as is, even if aws deprecates.

PR #518, opened 2019-05-14
Removing path style as option as it should only be used with non DNS …

tedPen:
…compliant buckets

As a result of policy changes by AWS, accessing s3 using the path_style will only be supported for buckets created on or before 11/30/2020. As such this change removes the path_style as a configuration option but leaves in place safeguards to convert invalid bucket names to path_style calls.

#516 change discussed here
[updated announcement](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/)

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -43,7 +43,7 @@ module Fog
       ]
 
       requires :aws_access_key_id, :aws_secret_access_key
-      recognizes :endpoint, :region, :host, :port, :scheme, :persistent, :use_iam_profile, :aws_session_token, :aws_credentials_expire_at, :path_style, :acceleration, :instrumentor, :instrumentor_name, :aws_signature_version, :virtual_host, :cname
+      recognizes :endpoint, :region, :host, :port, :scheme, :persistent, :use_iam_profile, :aws_session_token, :aws_credentials_expire_at, :acceleration, :instrumentor, :instrumentor_name, :aws_signature_version, :virtual_host, :cname
 
       secrets    :aws_secret_access_key, :hmac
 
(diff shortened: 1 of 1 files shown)


lib/fog/aws/storage.rb
```diff
@@ -278,15 +278,13 @@ def request_params(params)
...
-                if COMPLIANT_BUCKET_NAMES !~ bucket_name
-                  Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) is not a valid dns name, which will negatively impact performance.  For details see: http://docs.amazonwebservices.com/AmazonS3/latest/dev/BucketRestrictions.html")
-                  path_style = true
-                elsif scheme == 'https' && !path_style && bucket_name =~ /\./
-                  Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) contains a '.' so is not accessible over https as a virtual hosted bucket, which will negatively impact performance.  For details see: http://docs.amazonwebservices.com/AmazonS3/latest/dev/BucketRestrictions.html")
-                  path_style = true
-                end
+              path_style = false
+              if COMPLIANT_BUCKET_NAMES !~ bucket_name
+                Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) is not a valid dns name, which will negatively impact performance. For details see: http://docs.amazonwebservices.com/AmazonS3/latest/dev/BucketRestrictions.html")
```
geemus: This looks like a good approach based on our discussion. Maybe we should tweak the warning messages to refer to deprecation and link to the resources we found though? Or maybe even just link to the issue where we discussed this so people can get context (especially since it has the other links). What do you think?

geemus:
@tedPen Thanks for the extra details, digging up the updated policy on deprecation, and sharing the links. Seems like we are still on the right track here. Do you have thoughts on how best to test this?

lib/fog/aws/storage.rb
tedPen: I was thinking about this too, the existing link does give details on requirements for bucket names. But I think linking to the github issue is probably the best option.

geemus: Sounds good to me. We can certainly add a reference in the issue to this bucket restrictions page for reference also.

tillepille:
I would strongly disagree with this behavior! 
Even though AWS ditches the path variant, things like minio and other cloud providers which have a S3 compatible service do not.

geemus:
@tillepille Good callout, I must have overlooked that angle at the time I was reviewing this. I'll go ahead and close this based on ensuring we don't break the compatibility. Thanks!

Issue #520, opened 2019-05-18
Fog::AWS::Glacier::TreeHash::add_part works only sometimes


hkmaly:
You can easily test that from following three codes, only two returns same (presumably correct) result. And it would get even worse if the parts are not multiples of megabyte. Now, I suppose code using this internally is written so it always works correctly ... if the part size is correct ... and the input io doesn't do some weird buffering ... but in my opinion, the code should either work correctly or raise exception if some of those assumptions are broken, and never return incorrect value.
```
tree_hash = Fog::AWS::Glacier::TreeHash.new
tree_hash.add_part('x' * 1024*1024)
tree_hash.add_part('y' * 1024*1024)
tree_hash.add_part('z' * 1024*1024)
tree_hash.add_part('t' * 1024*1024)
tree_hash.hexdigest
```
```
tree_hash = Fog::AWS::Glacier::TreeHash.new
tree_hash.add_part( ('x' * 1024*1024) + ('y' * 1024*1024) + ('z' * 1024*1024))
tree_hash.add_part('t' * 1024*1024)
tree_hash.hexdigest
```
```
tree_hash = Fog::AWS::Glacier::TreeHash.new
tree_hash.add_part( ('x' * 1024*1024) + ('y' * 1024*1024) + ('z' * 1024*1024) + ('t' * 1024*1024))
tree_hash.hexdigest
```

geemus:
I've never personally used this (or glacier for that matter), so unfortunately I'm not super familiar with expectations around any of it. Maybe we could start with adding some failing stuff to https://github.com/fog/fog-aws/blob/master/tests/requests/glacier/tree_hash_tests.rb ? That way it's abundantly clear what expectations are, and where they are breaking. What do you think?

hkmaly:
I think I will make a pull request for this one. I've got idea how to fix it which was so surprisingly easy (relatively) I failed to resist temptation to try that. Also, unlike #519, this one couldn't cost money to test.

geemus:
Sounds good, thanks!

PR #521, opened 2019-05-21
Add support for generating tree hash tests by adding unaligned parts.

hkmaly:
As discussed in #520 the previous Fog::AWS::Glacier::TreeHash::add_part works only in very specific and undocumented circumstances which are luckily exactly the ones actually happening when using the gem.
However, it's not that hard to make it work truly generally, replacing the list of digests by stack with one digest for every power of two and additional digest object with unfinished last chunk.
As additional bonus, it will reduce the amount of memory consumed when processing large files, although I doubt it will be relevant unless it will go into petabytes.

### DIFF
--- a/tests/requests/glacier/tree_hash_tests.rb
+++ b/tests/requests/glacier/tree_hash_tests.rb
@@ -59,4 +59,26 @@ Shindo.tests('AWS::Glacier | glacier tree hash calcuation', ['aws']) do
 
   end
 
+  tests('multipart with unaligned parts') do
+    tree_hash = Fog::AWS::Glacier::TreeHash.new
+    part = ('x' * 512*1024)
+    returns(Fog::AWS::Glacier::TreeHash.digest(part)) { tree_hash.add_part part }
+
+    part2 = ('x' * 512*1024) + ('y'*1024*1024) + ('z'* 512*1024)
+    returns(Fog::AWS::Glacier::TreeHash.digest(part + part2)) { tree_hash.add_part part2 ; tree_hash.hexdigest }
+
+    tree_hash.add_part('z'* 512*1024 + 't'*1024*1024)
+
+    expected = OpenSSL::Digest::SHA256.hexdigest(
+                 OpenSSL::Digest::SHA256.digest(
+                    OpenSSL::Digest::SHA256.digest('x' * 1024*1024) + OpenSSL::Digest::SHA256.digest('y' * 1024*1024)
+                 ) +
+                 OpenSSL::Digest::SHA256.digest(
+                   OpenSSL::Digest::SHA256.digest('z' * 1024*1024) + OpenSSL::Digest::SHA256.digest('t' * 1024*1024)
+                 )
+               )
+    returns(expected) { tree_hash.hexdigest}
+
+  end
+
 end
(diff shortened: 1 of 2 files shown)


hkmaly:
Note: for higher compatibility, Fog::AWS::Glacier::TreeHash.new.hexdigest (meaning: calling the digest before adding any part) will crash just like previous version :-)

tests/requests/glacier/tree_hash_tests.rb
```diff
@@ -59,4 +59,26 @@
 
   end
 
+  tests('multipart with unaligned parts') do
```
geemus: It took me pondering this for a bit to understand what you meant by unaligned. So I think it might be good (for me and for posterity) if we were more explicit. If I understand, the issue is "multipart with parts less than one megabyte" that causes the issue. My initial guess was that having some x's in the first part and the rest in the second part was the issue, but if I'm not mistaken, it would also fail if the first 512*1024 were 'a' and the second were 'b', in the same way. Does that make sense? Is my understanding accurate?

hkmaly: Unaligned means not starting on (one megabyte) boundary, that is not starting on offset which is multiple of one megabyte. Obviously, only way for some parts to not start on one megabyte boundary is when some previous part is not ending on one, but if it's **last** part not ending on the boundary the original code would already work.

It's true that if first part would be 512*1024*'a' and then you would add second part with some b's it would be same problem as the b's would be unaligned. However, you can reproduce the problem also with all parts bigger than one megabyte - if the first part would be 1.5MB (second is 2MB and third is 1.5MB), all parts would be bigger than one megabyte but only the first would be aligned because neither 1.5MB nor 3.5MB is multiple of 1MB.

Note that one megabyte is only the **smallest** size exhibiting the problem. Original code would also hit problem if you would add a 2MB part not aligned on 2MB (not starting on 2MB boundary).

It's the same sense of "align" as described on https://en.wikipedia.org/wiki/Data_structure_alignment - except we are not dealing with data in memory, but with parts in "virtual" space of whole file.

As someone who started programming on C and assembler, this is obvious to me, but it's true you don't deal with it so often in higher languages so maybe it's not so clear nowadays?

hkmaly: Now, regarding the code: the most important part to understand is the "stack" array. Stack is again term from low level programming, but that's not important here. Important is that first position in that array corresponds with 1MB, second 2MB, third 4MB, fourth 8MB and so on. In any time, the size of all already added parts is equal to sum of all existing (non-nil) positions multiplied by that number, plus last_chunk_length for the remainder smaller than one megabyte. So, if last_chunk_length is half megabyte, stack[0] is filled, stack[1] and stack[2] empty and stack[3] filled, the size is 0.5MB + 1*1MB + 0*2MB + 0*4MB + 0*8MB = 9.5MB.

I guess this part would **definitely** warrant explanation comment in code.

geemus: Thanks for the extra explanation. Yeah, I think if you could add some explanation comments that might be helpful. I'm pretty sure I now understand, but I'm worried others might not (and/or when I come back to it later I won't recall). Thanks!

hkmaly:
Added some documentation. It might be too much but on the other hand, lot of people would dispute the idea that it's possible to have too much documentation.

geemus:
Thanks for working through this with me, looks good!

PR #522, opened 2019-05-30
Add eu north 1

dtaberer:


geemus:
I believe eu-north has already been added on master? It looks like it exists here: https://github.com/fog/fog-aws/blob/master/lib/fog/aws.rb#L228 (we added it about a month ago). Please do let me know if I'm missing something though. Thanks!

PR #524, opened 2019-06-07
Fixed issue with InvocationType header for AWS Lambda

GarrisonD:
According to [documentation](https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html) we should use `X-Amz-Invocation-Type` instead of `X-Amz-Client-Type`.


GarrisonD:
@geemus Are you going to release a new version soon? I need this fix in production 🙂 

I know that I can install this gem from GitHub but I would prefer to install it from `rubygems.org`

geemus:
Released as part of v3.5.1. Thanks!

PR #525, opened 2019-06-22
WIP: Make S3 Signature v4 streaming optional

stanhu:
Right now fog-aws will disable streaming for both multipart and PUT object transfers. However, Oracle Cloud appears to handle streaming fine in the multipart upload case; it's just the single PUT object case where it fails.


stanhu:
@geemus Would you mind tagging a new release? Thank you!

geemus:
Released in v3.5.2.

PR #530, opened 2019-08-14
Fix VPC tenancy on creation

ramonpm:
**Bug:** 
When you try to create a VPC with tenancy set to dedicated like this:
`ec2.vpcs.create(cidr_block: request.cidr, tenancy: 'dedicated')`
the tenancy is set to "default" when the creation request is done.

**RC:**
VPC#save was not passing to create_vpc request the InstanceTenancy option.

**Fix:**
- Changed #save method to pass InstanceTenancy argument.
- Corrected mock response and added a test case.

### DIFF
--- a/lib/fog/aws/models/compute/vpc.rb
+++ b/lib/fog/aws/models/compute/vpc.rb
@@ -106,7 +106,13 @@ module Fog
 
         def save
           requires :cidr_block
-          data = service.create_vpc(cidr_block, { 'AmazonProvidedIpv6CidrBlock' => amazon_provided_ipv_6_cidr_block }).body['vpcSet'].first
+
+          options = {
+            'AmazonProvidedIpv6CidrBlock' => amazon_provided_ipv_6_cidr_block,
+            'InstanceTenancy' => tenancy
+          }
+
+          data = service.create_vpc(cidr_block, options).body['vpcSet'].first
           new_attributes = data.reject {|key,value| key == 'requestId'}
           new_attributes = data.reject {|key,value| key == 'requestId' || key == 'tagSet' }
           merge_attributes(new_attributes)
--- a/lib/fog/aws/requests/compute/create_vpc.rb
+++ b/lib/fog/aws/requests/compute/create_vpc.rb
@@ -11,7 +11,6 @@ module Fog
         # * options<~Hash>:
         #   * InstanceTenancy<~String> - The allowed tenancy of instances launched into the VPC. A value of default
         #     means instances can be launched with any tenancy; a value of dedicated means instances must be launched with tenancy as dedicated.
-        #     please not that the documentation is incorrect instanceTenancy will not work while InstanceTenancy will
         #
         # === Returns
         # * response<~Excon::Response>:
@@ -54,7 +53,8 @@ module Fog
                 'classicLinkEnabled'          => false,
                 'classicLinkDnsSupport'       => false,
                 'cidrBlockAssociationSet'     => [{ 'cidrBlock' => cidrBlock, 'state' => 'associated', 'associationId' => "vpc-cidr-assoc-#{vpc_id}" }],
-                'ipv6CidrBlockAssociationSet' => []
+                'ipv6CidrBlockAssociationSet' => [],
+                'instanceTenancy'             => options['InstanceTenancy'] || 'default'
               }
               self.data[:vpcs].push(vpc)
 
--- a/tests/requests/compute/vpc_tests.rb
+++ b/tests/requests/compute/vpc_tests.rb
@@ -68,6 +68,12 @@ Shindo.tests('Fog::Compute[:aws] | vpc requests', ['aws']) do
       data
     end
 
+    tests("#create_vpc('10.255.254.0/28', {'InstanceTenancy' => 'dedicated'})").returns('dedicated') do
+      data = Fog::Compute[:aws].create_vpc('10.255.254.0/28',
+        {'InstanceTenancy' => 'dedicated'}).body
+      data['vpcSet'].first['instanceTenancy']
+    end
+
     tests('#describe_vpcs').formats(@describe_vpcs_format) do
       Fog::Compute[:aws].describe_vpcs.body
     end


lib/fog/aws/requests/compute/create_vpc.rb
```diff
@@ -11,7 +11,6 @@ class Real
         # * options<~Hash>:
         #   * InstanceTenancy<~String> - The allowed tenancy of instances launched into the VPC. A value of default
         #     means instances can be launched with any tenancy; a value of dedicated means instances must be launched with tenancy as dedicated.
-        #     please not that the documentation is incorrect instanceTenancy will not work while InstanceTenancy will
```
geemus: Is this no longer true? Do we need to worry about backwards compatibility issues?

lib/fog/aws/models/compute/vpc.rb
```diff
@@ -106,7 +106,13 @@ def disable_classic_link_dns
 
         def save
           requires :cidr_block
-          data = service.create_vpc(cidr_block, { 'AmazonProvidedIpv6CidrBlock' => amazon_provided_ipv_6_cidr_block }).body['vpcSet'].first
+
+          options = {
+            'AmazonProvidedIpv6CidrBlock' => amazon_provided_ipv_6_cidr_block,
+            'InstanceTenancy' => tenancy
```
geemus: There is an alias for the `instanceTenancy` above, do we need to alias `InstanceTenancy` as well?

lib/fog/aws/requests/compute/create_vpc.rb
ramonpm: It's true that lowercase won't work, but it's not true that the documentation is incorrect anymore. That link to documentation shows Capital I too, hence no problem with backward compatibility either. It was just a documentation update on their end.

lib/fog/aws/models/compute/vpc.rb
ramonpm: saw that alias and didn't really understand the use of it, could you please show me what you're thinking?

Basically what I'm trying to do here is pass the argument to create_vpc, we don't need anything beyond that.

lib/fog/aws/requests/compute/create_vpc.rb
geemus: Got it, thanks for clarifying. Sounds good.

lib/fog/aws/models/compute/vpc.rb
geemus: The value you are getting here from `tenancy` originates from that alias. Basically if you create a new model, when you pass in attributes, `InstanceTenancy` will get translated into the `tenancy` value due to that alias. But if you created a new vpc with `InstanceTenancy`, I'm not certain it would still set tenancy. Does that clarify?

ramonpm: understood, and it certainly wouldn't
```
[17] pry(main)> Fog::AWS::Compute::VPC.new('instanceTenancy' => 'dedicated')
=>   <Fog::AWS::Compute::VPC
    id=nil,
    state=nil,
    cidr_block=nil,
    dhcp_options_id="default",
    tags=nil,
    tenancy="dedicated",
    is_default=nil,
    cidr_block_association_set=nil,
    ipv6_cidr_block_association_set=nil,
    amazon_provided_ipv6_cidr_block=false
  >
[18] pry(main)> Fog::AWS::Compute::VPC.new('InstanceTenancy' => 'dedicated')
=>   <Fog::AWS::Compute::VPC
    id=nil,
    state=nil,
    cidr_block=nil,
    dhcp_options_id="default",
    tags=nil,
    tenancy="default",
    is_default=nil,
    cidr_block_association_set=nil,
    ipv6_cidr_block_association_set=nil,
    amazon_provided_ipv6_cidr_block=false
  >
[19] pry(main)> 
```
However, it doesn't matter as the **response** is lower case, accordingly to the documentation at least https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Vpc.html
Right now my account is returning `Fog::AWS::Compute::Error: VpcLimitExceeded => The maximum number of VPCs has been reached.` so I can't really test this on Real.
I'll test as soon as I can clean some VPC which will take time, or maybe if you can test it yourself...

geemus: If the response is lowercase it should probably be safe. If you are able to verify that would be great, I haven't actually directly run anything on AWS in ages (not even sure I still have valid creds at hand).

geemus: FWIW, I think many of the AWS limits are soft-ish. ie if you have a reasonable use case/reason and ask nicely they can often bump limits for you.

ramonpm:
Got it tested on Real =).

image

```
[3] pry(main)> vpc = ec2.vpcs.create(cidr_block: '10.0.0.1/16', tenancy: 'dedicated')
=>   <Fog::AWS::Compute::VPC
    id="vpc-01499bc336d6e2876",
    state="associated",
    cidr_block="10.0.0.0/16",
    dhcp_options_id="dopt-566b5a2e",
    tags=nil,
    tenancy="dedicated",
    is_default=nil,
    cidr_block_association_set=nil,
    ipv6_cidr_block_association_set=nil,
    amazon_provided_ipv6_cidr_block=false
  >
[4] pry(main)> vpc.wait_for { ready? }
=> {:duration=>0}
[5] pry(main)> vpc
=>   <Fog::AWS::Compute::VPC
    id="vpc-01499bc336d6e2876",
    state="available",
    cidr_block="10.0.0.0/16",
    dhcp_options_id="dopt-566b5a2e",
    tags={},
    tenancy="dedicated",
    is_default=false,
    cidr_block_association_set=[{"cidrBlock"=>"10.0.0.0/16", "associationId"=>"vpc-cidr-assoc-095481d9be28a67bb", "state"=>"associated"}],
    ipv6_cidr_block_association_set=[],
    amazon_provided_ipv6_cidr_block=false
  >
[6] pry(main)> vpc = ec2.vpcs.get('vpc-01499bc336d6e2876')
=>   <Fog::AWS::Compute::VPC
    id="vpc-01499bc336d6e2876",
    state="available",
    cidr_block="10.0.0.0/16",
    dhcp_options_id="dopt-566b5a2e",
    tags={},
    tenancy="dedicated",
    is_default=false,
    cidr_block_association_set=[{"cidrBlock"=>"10.0.0.0/16", "associationId"=>"vpc-cidr-assoc-095481d9be28a67bb", "state"=>"associated"}],
    ipv6_cidr_block_association_set=[],
    amazon_provided_ipv6_cidr_block=false
  >
```

geemus:
Awesome, thanks for working through that with me. I just wanted to catch it before accepting/releasing if possible. Will bring this in. Thanks!

ramonpm:
Sure, always a pleasure.

Issue #377, opened 2017-07-18
Support S3 object tagging


toupeira:
See http://docs.aws.amazon.com/AmazonS3/latest/dev/object-tagging.html

Looking at the code it looks like only bucket tags are implemented. Adding full support for object tags would involve at least the following:

- supporting the `x-amz-tagging` request header on [PUT Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html)
- supporting the `x-amz-tagging-count` response header on [GET Object](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html)
- supporting the [GET Object tagging](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGETtagging.html), [PUT Object tagging](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUTtagging.html) and [DELETE Object tagging](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectDELETEtagging.html) methods

geemus:
Sounds good, would you have availability to help us with a PR toward this? Thanks!

toupeira:
@geemus unfortunately no, for our use-case we ended up using directories for grouping.

geemus:
@toupeira ah, fair enough. Should that change in the future I would certainly welcome the help. In the mean time it will probably wait for somebody with a strong enough need/desire mix to work through it. Thanks for the update.

PR #537, opened 2019-11-07
Add object tagging

gustavosobral:
What is done:

- I added the `tags` attribute to the `Storage::File` model, so at creation time we can set the Object tags.
- I added the `get_object_tagging` and `put_object_tagging` requests to the `Storage`.

This PR partially solves #377.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 mkmf.log
 tests/.fog
 gemfiles/Gemfile-edge.lock
+.idea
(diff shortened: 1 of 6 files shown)


cryokev:
If any body just got here and wondering how to an object tag, let me save you some time:

```
aws = Fog::Storage.new({
                                provider: 'AWS',
                                aws_access_key_id: 'foo',
                                aws_secret_access_key: 'bar',
                                region: 'region'
                            })

aws.put_object_tagging('bucket_name', 's3_object_key', {hello: 'world'})
```

PR #549, opened 2020-02-26
Fixes type in class name for STS assume_role_with_web_identity parser

### LINKED ISSUE #548: Class name typo: AssumeRoleWithWithWebIdentity
In `lib/fog/aws/parsers/sts/assume_role_with_web_identity.rb` the class is named `AssumeRoleWithWithWebIdentity`, is that a typo?

Should it be named `AssumeRoleWithWebIdentity` instead? (Note WithWith instead of With)
as it's used here: `lib/fog/aws/requests/sts/assume_role_with_web_identity.rb`

dgoradia:
Fixes #548


geemus:
Thanks!

PR #550, opened 2020-02-27
Adds missing param WebIdentityToken for the request to the AWS api

dgoradia:
Fixes `Fog::AWS::STS::ValidationError (1 validation error detected: Value null at 'webIdentityToken' failed to satisfy constraint: Member must not be null)`


geemus:
Thanks again, and sorry for the merge-delay, had a bit of a hectic week with some work travel.

PR #552, opened 2020-03-13
Adding two missing regions to Fog::AWS.regions

lvangool:
Added the following to `lib/fog/aws.rb`
**me-south-1** - Middle East (Bahrain)
**us-gov-east-1** - AWS GovCloud (US) East

**NOTES**
1. **me-south-1** is not enabled by default for every AWS account and you might need to enable it before you can use this region
2. **us-gov-east-1** is part of AWS GovCloud (US)

### DIFF
--- a/lib/fog/aws.rb
+++ b/lib/fog/aws.rb
@@ -230,9 +230,11 @@ module Fog
         'eu-central-1',
         'eu-north-1',
         'eu-west-1', 'eu-west-2', 'eu-west-3',
+        'me-south-1',
         'us-east-1', 'us-east-2',
         'us-west-1', 'us-west-2',
         'sa-east-1',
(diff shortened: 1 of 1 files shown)


geemus:
Thanks!

lvangool:
@geemus No problem at all - thank you!

Would it be possible to cut a new version of the Gem? 
(Not sure if you do it on demand, or on a schedule)

geemus:
Hey, yes. It's generally more of an on demand thing these days. I don't think I'll get to it today, but I think I should be able to get a release done some time in the next few days. Will try to update here when I've had a chance to do so.

geemus:
Now released in v3.6.2.

lvangool:
@geemus Thank you so much! 👍

PR #553, opened 2020-03-26
Update kinesis.rb

ioquatix:
Ensure all data access is protected by mutex, and that the mutex is not lazy initialized.

### DIFF
--- a/lib/fog/aws/kinesis.rb
+++ b/lib/fog/aws/kinesis.rb
@@ -125,23 +125,26 @@ module Fog
       end
 
       class Mock
-        def self.mutex
-          @mutex ||= Mutex.new
-        end
-        def mutex; self.class.mutex; end
+        @mutex = Mutex.new
 
         def self.data
-          @data ||= Hash.new do |hash, region|
-            hash[region] = Hash.new do |region_hash, key|
-              region_hash[key] = {
-                :kinesis_streams => {}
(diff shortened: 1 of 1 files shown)


geemus:
Thanks for the patch. I think I understand why we would want to wrap the data access in the mutex (and not just the sequence number stuff). I'm a little less sure on the lazy init part though. Was that leading to a particular problem? Otherwise, I worry that it would lead to a lot of people now initializing a mutex that wouldn't get used (if they weren't using mocks, and/or weren't using kinesis for example). Does that make sense?

ioquatix:
Yes, it makes sense.

You have two options.

- If you still want to lazy initialize the mutex, you need to use a global mutex to lock around the lazy initialisation which will increase contention.

- Alternatively you could move the code to a different file so it's not loaded in a production environment where it's not needed.

At the end of the day I'm advising you that the code has an issue, and showing you a potential fix. It's up to you to implement what you think makes sense. This was not causing a particular problem. I am just analysing thread safety issues in Ruby code.

geemus:
Thanks for the additional context. I suspect a single, unused mutex won't actually be too much of a problem for users (and it allows us to keep this simple). So I'll bring this in, and we can revisit if/when it becomes needed. Thanks!

ioquatix:
Great!

PR #557, opened 2020-04-24
Don’t install development scripts

amarshall:
The scripts in `bin` appear intended for development, and have very generic names. They should not be installed as executables when installing the gem. Instead, change such executables to come from `exe`, which is more common.


amarshall:
Thanks for merging, it would be great to see a release with this soon as it resolves a compatibility issue with other software.

geemus:
released in 3.6.4

PR #569, opened 2020-06-18
hibernation option to compute

taniahagan:
Add the ability to enable hibernation on instances and activate hibernation


lib/fog/aws/requests/compute/stop_instances.rb
```diff
@@ -16,9 +16,13 @@ class Real
         #     * TODO: fill in the blanks
         #
         # {Amazon API Reference}[http://docs.amazonwebservices.com/AWSEC2/latest/APIReference/ApiReference-query-StopInstances.html]
-        def stop_instances(instance_id, force = false)
+        def stop_instances(instance_id, options = {})
           params = Fog::AWS.indexed_param('InstanceId', instance_id)
-          params.merge!('Force' => 'true') if force
+          params.merge!('Force' => 'true') if options['force']
```
geemus, line 21: For backwards compatibility, could you tweak this to check to see if options is a boolean value, and if so print a warning and then automatically convert it to the appropriate hash? I think that should strike a good balance of adding the new functionality while also preventing a breaking change for existing users. Does that make sense?

taniahagan: It does, I shall take a look at it

taniahagan: Ok, added the change

geemus: Looks great, thanks!

lib/fog/aws/requests/compute/run_instances.rb
```diff
@@ -221,6 +232,7 @@ def run_instances(image_id, min_count, max_count, options = {})
               'associatePublicIP'     => options['associatePublicIP'] || false,
               'architecture'          => 'i386',
               'blockDeviceMapping'    => block_device_mapping,
+              'hibernationOptions'    => hibernation_options,
```
geemus, line 235: Hey, I'm not sure why the integration isn't working right, but travis-ci is reporting an error here: https://travis-ci.org/github/fog/fog-aws/jobs/699781152#L413. In particular, `hibernation_options` is referred to is this mock version of the method without having previously been defined. We may need to either refer directly to the options hash or do something similar to the delete above. Does that make sense? Thanks!

taniahagan: Hopefully this change should do it.  Tested successfully running local mock test.

geemus:
Thanks for working through that with me. Do you need a new release with these changes?

taniahagan:
A new release would be good, though if one is due soon happy to wait a while longer.

geemus:
Released in v3.6.6. Thanks!

taniahagan:
Excellent, thank you very much.

PR #572, opened 2020-08-03
Make it possible to pass in headers to head_object

stanhu:
For using AWS SSE-C, custom headers such as
`x-amz-server-side-encryption-customer-algorithm` and
`x-amz-server-side-encryption-customer-key` are needed to perform the
HEAD request.

### DIFF
--- a/lib/fog/aws/requests/storage/head_object.rb
+++ b/lib/fog/aws/requests/storage/head_object.rb
@@ -34,7 +34,7 @@ module Fog
           if version_id = options.delete('versionId')
             query = {'versionId' => version_id}
           end
-          headers = {}
+          headers = (options[:headers] || {}).dup
           headers['If-Modified-Since'] = Fog::Time.at(options['If-Modified-Since'].to_i).to_date_header if options['If-Modified-Since']
           headers['If-Unmodified-Since'] = Fog::Time.at(options['If-Unmodified-Since'].to_i).to_date_header if options['If-Modified-Since']
           headers.merge!(options)


lib/fog/aws/requests/storage/head_object.rb
```diff
@@ -34,7 +34,7 @@ def head_object(bucket_name, object_name, options={})
           if version_id = options.delete('versionId')
             query = {'versionId' => version_id}
           end
-          headers = {}
+          headers = (options[:headers] || {}).dup
           headers['If-Modified-Since'] = Fog::Time.at(options['If-Modified-Since'].to_i).to_date_header if options['If-Modified-Since']
           headers['If-Unmodified-Since'] = Fog::Time.at(options['If-Unmodified-Since'].to_i).to_date_header if options['If-Modified-Since']
           headers.merge!(options)
```
geemus, line 40: It's a little obscure, but I think this line actually could already get you what you desired.

I think for this case you could pass: `options['header-key'] = example` for the three params and they should get merged in as expected here. It's been a while since I've worked on/around these parts, but a brief look at get and put appear that they already include something to this effect, so I think this would work consistently.

Could you try that and see if perhaps that solves your problem? If so, it may be that this is more of a documentation issue than one of code. Even then, we should definitely make it clearer how to use the encryption parameters for you and others. Thanks for digging in to this!

stanhu: Ah, right, I think that's it. Thanks!

PR #573, opened 2020-08-10
Add missing flavor

toubs13:


geemus:
@toubs13 Hmm. This pull request seems to include an awfully lot in addition to the flavor addition. Could you clean it up to just focus on this particular change for us? Thanks!

PR #576, opened 2020-10-26
Add AssumeRoleWithWebIdentity to fetch_credentials

jpac-run:
The action mentioned on the PR title can be found here: [AssumeRoleWithWebIdentity](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html#API_AssumeRoleWithWebIdentity_Examples)

The `nokogiri` gem was added to parse the XML response from the host.

### DIFF
--- a/lib/fog/aws/credential_fetcher.rb
+++ b/lib/fog/aws/credential_fetcher.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "nokogiri"
-
 module Fog
   module AWS
     module CredentialFetcher
--- a/lib/fog/aws/credential_fetcher.rb
+++ b/lib/fog/aws/credential_fetcher.rb
@@ -1,3 +1,7 @@
+# frozen_string_literal: true
+
+require "nokogiri"
+
 module Fog
   module AWS
     module CredentialFetcher
@@ -8,6 +12,8 @@ module Fog
       INSTANCE_METADATA_AZ = "/latest/meta-data/placement/availability-zone/"
 
       CONTAINER_CREDENTIALS_HOST = "http://169.254.170.2"
+
+      STS_GLOBAL_ENDPOINT = "https://sts.amazonaws.com"
 
       module ServiceMethods
         def fetch_credentials(options)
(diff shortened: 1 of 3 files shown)


lib/fog/aws/credential_fetcher.rb
```diff
@@ -1,3 +1,7 @@
+# frozen_string_literal: true
+
+require "nokogiri"
```
geemus, line 3: I think this should already be available via `fog-xml`? Might be better to not have the explicit include here also unless we really need it?

geemus (changes requested):
One small question around requires, otherwise looks good. Thanks!

lib/fog/aws/credential_fetcher.rb
jpac-run: I've removed both the explicit require for `nokogiri` and gemfile dependency, and the tests are still green 👍 Did not check the `gemspec` file and the other `fog` gems.

geemus: No worries, it probably wouldn't hurt anything at the moment. But could come back to complicate things later, so probably better to fallback to the existing dependency to keep things simple. Thanks.

geemus (approved):

geemus:
Thanks!

lib/fog/aws/credential_fetcher.rb
```diff
@@ -23,6 +27,30 @@ def fetch_credentials(options)
...
+                if region.nil?
+                  connection = options[:metadata_connection] || Excon.new(INSTANCE_METADATA_HOST)
+                  token_header = fetch_credentials_token_header(connection, options[:disable_imds_v2])
+                  region = connection.get(:path => INSTANCE_METADATA_AZ, :idempotent => true, :expects => 200, :headers => token_header).body[0..-2]
+                end
+              elsif ENV["AWS_WEB_IDENTITY_TOKEN_FILE"]
+                params = {
+                  :Action => "AssumeRoleWithWebIdentity",
+                  :RoleArn => options[:role_arn] || ENV.fetch("AWS_ROLE_ARN"),
+                  :RoleSessionName => options[:role_session_name] || ENV.fetch("AWS_ROLE_SESSION_NAME"),
```
camhine, line 41: Hey folks :wave:. 

Apologies if this is the wrong place to raise this, but I just ran this in a K8s cluster and tripped over this line here. I hadn't set `role_session_name` or `AWS_ROLE_SESSION_NAME`.

Would it make sense to have a default value to fall back on? Something like:

```suggestion
                  :RoleSessionName => options[:role_session_name] || ENV.fetch("AWS_ROLE_SESSION_NAME") { 'fog-aws' },
```

As far as I can tell `RoleSessionName` is really just an alias for the session and isn't as important as `RoleArn` and `WebIdentityToken`.

geemus: I'll start by saying I don't have much personal experience with usage, and that I'm open to discussing. It does look like `RoleSessionName` isn't that important when assuming, but I imagine it could be more important for auditing purposes? ie it could help distinguish between different times you might use something (or who was doing what), which always having a default might glaze over? Does that make sense or am I overthinking or misunderstanding things? I'm open to the possibility of a default, but I don't want that to result in a worse security or audit stance.

camhine: I hadn't thought about the auditing side of things :thinking: 

Reading the [assume-role-with-web-identity](https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role-with-web-identity.html) docs more carefully, I think you're right:

> Typically, you pass the name or identifier that is associated with the user who is using your application. That way, the temporary security credentials that your application will use are associated with that user.

So this value should map to the user/identity that the app is running as - the user/identity which assumed the role. 

For anyone else reading, in EKS when using [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html), this value should probably be the `serviceAccountName` associated with the pod. `AWS_ROLE_SESSION_NAME` isn't injected into the pod automatically like `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` are.

Thanks for taking the time to look @geemus. Much appreciated.

geemus: Yeah, that lines up with my reading of that part of the docs also. During usage you likely will almost never care, but not being able to refer back later could easily come back to bite you. Seems probably worth being more cautious and expecting user-supplied values I think. Happy to help!

Issue #577, opened 2020-11-16
File#copy does not support files above 5 GB


stanhu:
`Fog::AWS::Storage::File#copy` uses the single CopyObject API: https://github.com/fog/fog-aws/blob/ab288f29a0974d64fd8290db41080e5578be9651/lib/fog/aws/models/storage/file.rb#L100-L105

However, this does not use multipart uploads. As a result, files above 5 GB fail to transfer.

In addition, I noticed the AWS SDK uses 10 threads in parallel to speed up the transfer (https://github.com/aws/aws-sdk-ruby/blob/d2dc0213758da0aee100e485f2dfb090ffa7dbd5/gems/aws-sdk-s3/lib/aws-sdk-s3/multipart_file_uploader.rb#L133-L163). As a result, uploads via the SDK are significantly faster than fog-aws (e.g. 6 GB file takes 7 minutes as opposed to a minute with the `aws` CLI).

Certainly implementing a multi-threaded version that uses multipart uploads could be implemented, but is there a reason fog-aws doesn't just call out to the AWS SDK in the first place?

geemus:
@stanhu hey, thanks for the detailed report and checking in. Copy was implemented like that, I think largely because it was simple and seemed good-enough at the time. We certainly could think about an implementation more like what you suggest. 

I worry a bit about the additional complexity of managing threads, debugging issues, etc if we did it ourselves. As for using aws-sdk directly, the short answer is that `aws-sdk` is not a dependency of fog-aws and we are trying to keep the dependency list small/short. If someone needs/wants this functionality, I guess I might suggest they utilize the aws-sdk directly, rather than adding it for this in particular. Do you think we should advise people to use aws-sdk directly if they need this, or is it important to change this and/or add this capability here?

stanhu:
@geemus That's a good question. Probably documentation should suffice for now. However, setting aside the copy case, I was surprised by how dramatic basic upload performance was without multithreading for a 2 GB file:

|Case|Time|
|-----|----|
|fog-aws|57 s|
|aws CLI|9.6 s|

[...]

PR #578, opened 2020-11-18
Add support for multipart Fog::AWS::Storage::File#copy

stanhu:
This enables copying of files larger than 5 GB.

Closes https://github.com/fog/fog-aws/issues/577

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -331,12 +331,6 @@ module Fog
             part_tags << part_upload.body['ETag']
             current_pos = end_pos + 1
           end
-
-          if part_tags.empty? #it is an error to have a multipart upload with no parts
-            part_upload = service.upload_part_copy(directory.key, key, upload_id, 1, upload_part_options)
-            part_tags << part_upload.body["ETag"]
-          end
-
         rescue => e
           # Abort the upload & reraise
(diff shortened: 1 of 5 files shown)


lib/fog/aws/models/storage/file.rb
```diff
@@ -294,6 +308,44 @@ def multipart_save(options)
...
+            end_pos = [current_pos + self.multipart_chunk_size, self.content_length - 1].min
+            range = "bytes=#{start_pos}-#{end_pos}"
+
+            upload_part_options['x-amz-copy-source-range'] = range
+            part_upload = service.upload_part_copy(target_directory_key, target_file_key, upload_id, part_tags.size + 1, upload_part_options)
+            part_tags << part_upload.body['ETag']
+            current_pos = end_pos + 1
+          end
+
+          if part_tags.empty? #it is an error to have a multipart upload with no parts
```
stanhu, line 335: I think we can get rid of this section because this should never happen.

geemus (approved):
LGTM, thanks!

lib/fog/aws/models/storage/file.rb
geemus: Yeah, not sure. We can always remove it later (though that seems unlikely to actually happen, as we are apt to forget).

PR #579, opened 2020-11-18
Add multi-threaded support for File#copy

stanhu:
This improves transfer performance significantly. For a 6 GB transfer, instead of taking several minutes to copy, it now appears to take 36 seconds on a local EC2 instance.

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -1,5 +1,4 @@
 require 'fog/aws/models/storage/versions'
-require 'concurrent-ruby'
 
 module Fog
   module AWS
@@ -32,12 +31,32 @@ module Fog
         attribute :kms_key_id,          :aliases => 'x-amz-server-side-encryption-aws-kms-key-id'
         attribute :tags,                :aliases => 'x-amz-tagging'
 
-        class UploadPartData
-          attr_reader :part_number, :upload_options
+        UploadPartData = Struct.new(:part_number, :upload_options, :etag)
 
-          def initialize(part_number, upload_options)
-            @part_number = part_number
-            @upload_options = upload_options
+        class PartList
+          def initialize(parts = [])
+            @parts = parts
+            @mutex = Mutex.new
+          end
+
+          def push(part)
+            @mutex.synchronize { @parts.push(part) }
+          end
+
+          def shift
+            @mutex.synchronize { @parts.shift }
+          end
+
+          def clear!
+            @mutex.synchronize { @parts.clear }
+          end
+
+          def size
+            @mutex.synchronize { @parts.size }
+          end
+
+          def to_a
+            @mutex.synchronize { @parts.dup }
           end
         end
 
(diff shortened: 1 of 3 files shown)


geemus:
@stanhu Does this get closer to the other benchmark you were comparing against timing wise?

stanhu:
@geemus Yes, definitely. The transfer times track the `aws` CLI performance now for the copy case. We can extend this for the upload case as well.

lib/fog/aws/models/storage/file.rb
```diff
@@ -1,4 +1,5 @@
 require 'fog/aws/models/storage/versions'
+require 'concurrent-ruby'
```
geemus, line 2: Won't this need to be added to the Gemfile and other dependencies? I'm a little reticent to add dependencies, but this one looks well vetted (and doesn't appear to have sub-dependencies to contend with).

lib/fog/aws/models/storage/file.rb
```diff
@@ -315,22 +325,15 @@ def multipart_copy(options, upload_part_options, target_directory_key, target_fi
 
           # Store ETags of upload parts
           part_tags = []
+          pending = create_part_list(upload_part_options)
 
-          current_pos = 0
+          thread_count = options['concurrency'] || 10
```
geemus, line 330: I wonder if it would be safer to default to 1 to start, and then once we have vetted it a bit we could bump it up to a higher number? I was thinking this might simplify debugging/etc if anything comes up, while allowing people to opt-in to the other behavior. Not set on one way or the other, but wanted to check-in and see your feeling.

geemus (changes requested):
A couple small points to discuss added in comments. Thanks!

lib/fog/aws/models/storage/file.rb
stanhu: @geemus For some reason, I thought `concurrent-ruby` was already included, but now I see that is my local `Gemfile.lock`. 😄 

I got rid of this dependency with a thread-safe queue that is similar to the AWS SDK.

lib/fog/aws/models/storage/file.rb
stanhu: Done. 

I made `concurrency` an attribute rather than a parameter `options` because I worried that the option would be passed along the HTTP request by this or other Fog providers. What do you think?

geemus (approved):

geemus:
@stanhu Sounds like a good plan. Have you had a chance to try this latest iteration live and ensure it performs appropriately/etc?

stanhu:
@geemus Yes, my test script looks something like:

```ruby
require_relative 'lib/fog-aws'

# create a connection
connection = Fog::Storage.new(
  {
    provider: 'AWS',
    aws_access_key_id: 'REDACTED',
    aws_secret_access_key: 'REDACTED',
    region: 'us-west-2'
})

dir = connection.directories.new(key: 'MY-BUCKET')
f = dir.files.get('2GB.bin')
f.multipart_chunk_size = 10 * 1024 * 1024
f.concurrency = ENV['CONCURRENCY']
puts f.key
f.copy('MY-BUCKET', '/tmp/2GB.bin.copy')
```

With concurrency set to 10, I see this 2GB transfer finishing around 13 seconds.

With concurrency set to 1, this transfer takes around 1 minute and 10 seconds.

geemus:
@stanhu sounds good, just wanted to double check that we weren't relying solely on the mocks. Thanks!

stanhu:
Thanks! Would you mind cutting a new release? I was going to work on the upload case, but that will take a little longer.

geemus:
Certainly, released in v3.7.0. Thanks!

PR #587, opened 2021-01-08
Rewind pointer if file is eof on put_object mock

ekulz:
### What problem does this solve?
When passing a file object that has reached end-of-file to the `put_object` mock, an empty string is saved to the uploaded file body property. This is because subsequent `read` calls will return an empty string as there is no more to read. Sometimes code does some operations such as reading and validating a file before uploading to S3. The real service seems to be fine with the file being at EOF and still uploads the data, however I noticed the mocks are not

### How does this solve it?
Rewinds the pointer of the file object if the file is at EOF


geemus (approved):
Looks good, thanks for the detailed report and fix!

PR #588, opened 2021-01-09
Fix NameError in multipart copy

stanhu:
If an error occurred in a multipart copy, the following error would
show:

```
undefined local variable or method `error' for #<Fog::AWS::Storage::File:0x00007f98212f9d68>
Did you mean?  errors
```


geemus (approved):
Good catch, and thanks for the fix!

PR #589, opened 2021-01-10
Filter unknown UploadPartCopy paramaters

stanhu:
Previously attempting to use S3 server side encryption headers
(e.g. `x-amz-server-side-encryption`) would fail when attempting to use
`File#copy` because the method would pass these headers to the `UploadPartCopy`
(https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html)
API call. However, this would fail with an error:

```
<Code>InvalidArgument</Code>
<Message>x-amz-server-side-encryption header is not supported for this operation.</Message>
<ArgumentName>x-amz-server-side-encryption</ArgumentName>
<ArgumentValue>AES256</ArgumentValue>
```

This header can and should be used in the `CompleteMultipartUpload`
call. To support this, we filter out unknown headers in the
`UploadPartCopy`.

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -148,7 +148,8 @@ module Fog
           self.multipart_chunk_size = MIN_MULTIPART_CHUNK_SIZE * 2 if !multipart_chunk_size && self.content_length.to_i > MAX_SINGLE_PUT_SIZE
 
           if multipart_chunk_size && self.content_length.to_i >= multipart_chunk_size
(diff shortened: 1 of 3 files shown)


geemus (approved):
Looks good, thanks!

stanhu:
@geemus Thanks for the quick merge! Would you mind tagging a new release?

geemus:
Sure, I should be able to do that in the next day or two.

geemus:
Just released as part of v3.8.0. Thanks!

PR #590, opened 2021-01-14
Fix typo in CHANGELOG

stanhu:
paramaters -> parameters

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 **Merged pull requests:**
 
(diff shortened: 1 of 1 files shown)


geemus (commented):
Good catch. Unfortunately, I generate this directly from commit log and it looks like we inadvertently let that slip in there (so this fix would be overwritten next time I regenerate). So it's probably best to just leave the typo here, as I don't think it's worth rewriting git history to fix. Thanks for your help with the other fixes and noticing this though.

PR #593, opened 2021-02-09
Fixes domain name duplication in url

midhunkrishna:
When a bucket url like `bucket.s3.amazonaws.com` with bucket name `bucket`is present in params passed into equest_params method, it should not duplicate the subdomain.

**current behavior**: bucket.bucket.s3.amazonaws.com
**expected behavior**: bucket.s3.amazonaws.com

Fixes: https://github.com/fog/fog-aws/issues/594

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -298,6 +298,8 @@ module Fog
                 host = params.fetch(:cname, bucket_name)
               elsif path_style
                 path = bucket_to_path bucket_name, path
+              elsif host.start_with?("#{bucket_name}.")
+                # no-op
               else
                 host = [bucket_name, host].join('.')
               end
--- a/tests/models/storage/directory_tests.rb
+++ b/tests/models/storage/directory_tests.rb
@@ -1,4 +1,117 @@
 Shindo.tests("Storage[:aws] | directory", ["aws"]) do
+  tests('Fog::Storage[:aws]', "#request_params") do
+    def slice(hash, *args)
+      hash.select { |k, _| args.include?(k) }
+    end
+
+    instance = Fog::Storage[:aws]
+    method = instance.method(:request_params)
+
+    params = {bucket_name: 'profile-uploads', host: 'profile-uploads.s3.us-west-2.amazonaws.com'}
+    tests("given #{params}, request_params[:host]").returns("profile-uploads.s3.us-west-2.amazonaws.com") do
+      method.call(params)[:host]
+    end
+
+    params = {bucket_name: 'profile-uploads.johnsmith.net', cname: 'profile-uploads.johnsmith.net', virtual_host: true}
+    tests("given #{params}, request_params[:host]").returns("profile-uploads.johnsmith.net") do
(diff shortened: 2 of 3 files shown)


lib/fog/aws/storage.rb
```diff
@@ -298,7 +298,7 @@ def request_params(params)
                 host = params.fetch(:cname, bucket_name)
               elsif path_style
                 path = bucket_to_path bucket_name, path
-              else
+              elsif !host.include?(bucket_name)
```
geemus, line 301: I think this might have a few edge cases, if the bucketname happens to be part of the normal amazon aws url. For example, a bucket named `north` would match `s3.ap-northeast-1.amazonaws.com`, even if the host didn't have the bucket name in it. Maybe we should specifically check to see if the host starts with the bucket_name to avoid this? (there are only a very few cases where what you did wouldn't work, but better safe than sorry). Does that make sense?

geemus (changes requested):
Thanks, the redirects have been a bit of a pain as it seems to be somewhat inconsistent what exactly they return (and/or maybe it has just changed over the course of time). In any event, overall this seems like a reasonable way to fix it for the case you ran into without breaking it for other cases. I just suggest a little bit stricter matching in the else to avoid a narrow set of edge cases. If you could make that change I think we should be good to go. Thanks!

lib/fog/aws/storage.rb
midhun-at-vineti: Makes sense, let me make that change 👍

midhunkrishna:
@geemus The bug is in the way `request_params` method constructs the final host, also `request_params` method is a private method. 

I have modified the code like this:
```ruby
def request_params(options)
  # ...
  if params[:virtual_host]
    host = params.fetch(:cname, bucket_name)
  elsif path_style
    path = bucket_to_path bucket_name, path
  elsif host.start_with?("#{bucket_name}.")
    # no-op
  else
    host = [bucket_name, host].join('.')
  end
end
```

I have explored two options:

**A.** Write test case for the `Fog::Storage['aws']#get_bucket` method, and make `Fog::XML::Connection#request` method call raise permanently moved exception, which would start the rescue block, and finally will call the request_params(https://github.com/fog/fog-aws/blob/master/lib/fog/aws/storage.rb#L253) method. 

The problem with this is that, I am not sure how to mocks/stub method calls with Shindo.

**B.** Write a test case just for request_params method, being a private method, what is the practice of writing test cases for that method (in the context of this library)?

geemus:
Hmm. Good questions. I guess I would probably lean toward tests around the `request_params` method itself as this seems less brittle (and less complicated). We should be able to use `send(:request_params)` to call it even though it is private for test purposes.

geemus:
Also, the new request_params version you listed in your post looks good to me, thanks!

midhun-at-vineti:
@geemus Updated PR with review suggestions. 👍

geemus (approved):

geemus:
@midhun-at-vineti looks great, thanks for working through that with me. Do you need a release?

PR #595, opened 2021-02-09
Generate a default session name if one is not provided

stanhu:
When using IAM roles for Service Accounts (IRSA), annotated pods get a number of
environment variables:

1. `AWS_WEB_IDENTITY_TOKEN_FILE`
2. `AWS_ROLE_ARN`
3. `AWS_DEFAULT_REGION`
4. `AWS_REGION`

`AWS_ROLE_SESSION_NAME` is not usually included. The AWS SDK falls back
to using `default_name` as the default session name if
`AWS_ROLE_SESSION_NAME` is not provided. From
https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-role_session_name.html:

> If you don't provide this value, a session name is generated
> automatically if the profile assumes a role.

### DIFF
--- a/lib/fog/aws/credential_fetcher.rb
+++ b/lib/fog/aws/credential_fetcher.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'securerandom'
+
 module Fog
   module AWS
     module CredentialFetcher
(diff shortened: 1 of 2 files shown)


mthoretton:
Looking at [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#assume-role-provider) `role_session_name` is optional. Maybe it's the same here and we can just omit it if it's not set.

stanhu:
Note that I got `session_name` from the AWS SDK here: https://github.com/aws/aws-sdk-ruby/blob/7c9584bc6473100df9aec9333ab491ad4faeeca8/gems/aws-sdk-core/lib/aws-sdk-core/shared_config.rb#L217

geemus:
@stanhu It looks like this is an optional param that gets some generated value if left blank. I can see that the ruby-sdk chooses to supply a value, but was wondering what your thoughts were about if we would prefer this specific value vs the generated ones? I'm not sure what the generated ones look like, but having a variety might be more helpful in some cases than having it always be the same. What do you think?

stanhu:
It sounds like we can just leave it blank since AWS should be generating a session name on its own.

stanhu:
@geemus Yeah, I just generated a random one since according to https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/STS.html#assumeRole-property the session name is always required.

geemus:
@stanhu sounds good, thanks for the detailed report and fix.

geemus (approved):

PR #597, opened 2021-02-23
Fix handling of lowercased HTTP headers

stanhu:
Some S3 providers (e.g. Ceph) appear to send responses with lowercase
HTTP headers (e.g. content-length vs. Content-Length). This appears to
confuse fog-aws.

The attributes were not being loaded because the aliases did not contain
the lowercased version, and the attributes themselves use underscores
rather than dashes.

To fix this issue, we normalize the headers by fetching the value via
excon's case-insensitive lookup, storing the canonical key, and then
deleting other matches.

### DIFF
--- a/lib/fog/aws/models/storage/files.rb
+++ b/lib/fog/aws/models/storage/files.rb
@@ -17,6 +17,15 @@ module Fog
 
         model Fog::AWS::Storage::File
 
+        DASHED_HEADERS = %w(
+          Cache-Control
+          Content-Disposition
+          Content-Encoding
+          Content-Length
+          Content-MD5
+          Content-Type
+        ).freeze
+
         def all(options = {})
           requires :directory
           options = {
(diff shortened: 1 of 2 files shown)


geemus:
Thanks for the detailed report and fix. FWIW, I think it is technically excon case-insensitive lookup, rather than faraday, though they probably work almost identically (it just threw me off when I read that initially). I have one bit of other feedback which I'll put in a review momentarily. Thanks!

lib/fog/aws/models/storage/files.rb
```diff
@@ -116,6 +125,11 @@ def new(attributes = {})
         def normalize_headers(data)
           data.headers['Last-Modified'] = Time.parse(data.get_header('Last-Modified'))
           data.headers['ETag'] = data.get_header('ETag').gsub('"','')
+
+          DASHED_HEADERS.each do |header|
+            value = data.get_header(header)
+            data.headers[header] = value if value
```
geemus, line 131: Although I suspect it doesn't cause any problems at present, I'm a little worried about how this may result in data with two different capitalization of the same key in it. I could see how this might lead, for instance, to people then updating the wrong one and being rather confused. I suspect it may be a bit cleaner/safer if we tweak this so that if/when we remap things we also delete the existing value (though maybe this is awkward if we don't know the exact capitalization). Does that make sense? What do you think?

geemus (changes requested):

lib/fog/aws/models/storage/files.rb
stanhu: @eemus I updated this pull request to clean up the headers. I was surprised that Excon didn't do this for us.

stanhu:
@geemus Friendly ping. Could you take a look at this and cut a new release if it looks good to you? Thank you so much as usual.

lib/fog/aws/models/storage/files.rb
geemus: Data headers still returning case sensitive was an intentional (albeit, perhaps overly cautious) choice. The idea being that in some cases an end user might specifically care about or check things that relate to case-sensitivity, so we wanted to give back what we got, rather than a mutated version. So, we were trying to provide a raw view, but also make it easier to work with. Again, maybe overthinking it, but there was a reason.

geemus (approved):
Thanks for working through this with me, sorry for the odd and surprising behaviors along the way.

geemus:
@stanhu thanks for the ping. I had meant to get back to this and it had at least temporarily fallen off my radar.

geemus:
@stanhu released in v3.9.0, thanks again for the help and patience.

stanhu:
Thanks as usual @geemus!

Issue #598, opened 2021-03-08
S3 multiple delete api should handle delete for multiple versions for a single object


shanu-idrive:
I am trying to use bulk delete API for S3. But FOG [implementation](https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/delete_multiple_objects.rb) has a limitation. The limitation is though it can delete
multiple objects but only 1 version of each object at a time. Further observing the code I found that it is due to the way it expects version_ids in options hash which should be key value pair for `object_name` and `version_id`. 

We expect the API to handle delete for multiple versions of single file as well. This can reduce the number of calls to delete the objects versions. Maybe a array of versions in versions_ids hash can be used. 
For e.g
objects:  `["folder/", "folder/file1","folder/file2"]`
options = `{ :quiet => true, 'versionId' => {"folder/"=> ["1614921964.48870"], "folder/file1"=> ["1614922015.63559", "1614921991.95230"], "folder/file2"=> ["1614922016.84576"]} }`

Or if there already exists a solution please let me know. 
And if it is missing feature I can raise a PR for this.

geemus:
Thanks for the detailed report.

Looking at the code, the comment seems to imply something kind of like what you are suggesting, but it looks like the code expects string versions (not arrays), just as you suggest.

I think the way that you suggest of allowing it to be arrays makes good sense. We'll want to make sure the documentation is clear, and I think we should make it so that if a string is passed it can turn it into an array inside the method automatically (for backwards compatibility). Let me know if that doesn't make sense or you have questions.

I'm pretty sure from your description and reviewing this that it must be a missing feature. If you would like to work on a PR to address it, that would be great. Thanks!

PR #600, opened 2021-03-17
Handle multiple versions of objects in multiple delete request

shanu-idrive:
This PR is fix for issue describe in #598 . Now bulk/multiple delete can handle multiple versions of single object as well in delete 
request. The solution is as discussed in the issue where options hash accepts versions as an array of strings for a object. The PR is backward compatible to support hash with versions as string only.

Thanks for your time.

### DIFF
--- a/lib/fog/aws/requests/storage/delete_multiple_objects.rb
+++ b/lib/fog/aws/requests/storage/delete_multiple_objects.rb
@@ -36,13 +36,20 @@ module Fog
           data << "<Quiet>true</Quiet>" if headers.delete(:quiet)
           version_ids = headers.delete('versionId')
           object_names.each do |object_name|
-            data << "<Object>"
-            data << "<Key>#{CGI.escapeHTML(object_name)}</Key>"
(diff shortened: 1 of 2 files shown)


geemus (approved):
Looks good, thanks for taking time to talk and work through this with me.

geemus:
@shanu-idrive Do you need a release with this?

shanu-idrive:
> @shanu-idrive Do you need a release with this?

@geemus thanks for you time for reviewing and approving the changes. Yes I would need a release as we are using this API in our product.

jgigault:
That's awesome! Thanks for this improvement

geemus:
Sounds good. I'll try to release tomorrow (don't want to do it on the weekend when I'm not around/available as much, just in case something goes wrong).

shanu-idrive:
> Sounds good. I'll try to release tomorrow (don't want to do it on the weekend when I'm not around/available as much, just in case something goes wrong).

Sure, as per your convenience.

geemus:
Should now be available as part of v3.10.0. Thanks!

PR #602, opened 2021-03-26
drop git in gemspec

abrahamparayil:


geemus (changes requested):
Thanks, that definitely will remove some unnecessary files. I do wonder if we might want to still include some of the other top-level files.

Basically the various ALLCAPS files like license and README seem like they would still be a good inclusion. Could you include those as well? Maybe something like: https://github.com/excon/excon/blob/master/excon.gemspec#L15

Thanks!

geemus (approved):

geemus:
Thanks!

PR #605, opened 2021-05-03
Add support for regional STS endpoints

### LINKED ISSUE #604: Support for Regional STS Endpoints
based on https://github.com/fog/fog-aws/blob/f381dd7166d048934d22cb4480235e9f311566f1/lib/fog/aws/credential_fetcher.rb#L16 and  https://github.com/fog/fog-aws/blob/f381dd7166d048934d22cb4480235e9f311566f1/lib/fog/aws/credential_fetcher.rb#L47 

It would seem fog-aws cannot currently support regional STS endpoints. This is a deal breaker in organizations that lock down global endpoints in favor of regional endpoints where they exist.

fog-aws should have logic that handles whether or not the environment variable "AWS_STS_REGIONAL_ENDPOINTS" exists with the value "regional" and is matched with the value of "AWS_REGION".

MichaelGerhart:
Fixes #604

### DIFF
--- a/lib/fog/aws/credential_fetcher.rb
+++ b/lib/fog/aws/credential_fetcher.rb
@@ -13,8 +13,6 @@ module Fog
 
       CONTAINER_CREDENTIALS_HOST = "http://169.254.170.2"
 
-      STS_GLOBAL_ENDPOINT = "https://sts.amazonaws.com"
-
       module ServiceMethods
         def fetch_credentials(options)
           if options[:use_iam_profile] && Fog.mocking?
@@ -44,7 +42,14 @@ module Fog
                   :WebIdentityToken => File.read(options[:aws_web_identity_token_file] || ENV.fetch("AWS_WEB_IDENTITY_TOKEN_FILE")),
                   :Version => "2011-06-15",
                 }
-                connection = options[:connection] || Excon.new(STS_GLOBAL_ENDPOINT, :query => params)
+
+                if ENV["AWS_STS_REGIONAL_ENDPOINTS"] == "regional" && ENV["AWS_DEFAULT_REGION"]
+                  STS_ENDPOINT = "https://sts.#{ENV['AWS_DEFAULT_REGION']}.amazonaws.com"
+                else
+                  STS_ENDPOINT = "https://sts.amazonaws.com"
+                end
+
+                connection = options[:connection] || Excon.new(STS_ENDPOINT, :query => params)
                 document = Nokogiri::XML(connection.get(:idempotent => true, :expects => 200).body)
 
                 session = {
--- a/tests/credentials_tests.rb
+++ b/tests/credentials_tests.rb
@@ -83,6 +83,7 @@ Shindo.tests('AWS | credentials', ['aws']) do
         aws_secret_access_key: 'dummysecret',
         aws_session_token: 'dummytoken',
         region: 'us-west-1',
+        STS_ENDPOINT: "https://sts.amazonaws.com"
         aws_credentials_expire_at: expires_at
       ) { Fog::AWS::Compute.fetch_credentials(use_iam_profile: true) }
     end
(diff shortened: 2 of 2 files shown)


geemus (changes requested):
Thanks for taking a stab at this, looks like we are headed in the right direction. I'll add a few specific comments and questions presently.

tests/credentials_tests.rb
```diff
@@ -83,6 +83,7 @@
         aws_secret_access_key: 'dummysecret',
         aws_session_token: 'dummytoken',
         region: 'us-west-1',
+        STS_ENDPOINT: "https://sts.amazonaws.com"
```
geemus, line 86: The all-caps here seems out of place compared to the other keys. Could we use lowercase for consistency?

tests/credentials_tests.rb
```diff
@@ -95,10 +96,27 @@
...
         aws_credentials_expire_at: expires_at
       ) { Fog::AWS::Compute.fetch_credentials(use_iam_profile: true) }
     end
 
+    ENV["AWS_STS_REGIONAL_ENDPOINTS"] = "regional"
+    ENV["AWS_REGION"] = "us-west-1"
```
geemus, line 105: I think, given the rest of the implementation details, that this should maybe use the `AWS_DEFAULT_REGION` key?

lib/fog/aws/credential_fetcher.rb
```diff
@@ -44,7 +42,14 @@ def fetch_credentials(options)
                   :WebIdentityToken => File.read(options[:aws_web_identity_token_file] || ENV.fetch("AWS_WEB_IDENTITY_TOKEN_FILE")),
                   :Version => "2011-06-15",
                 }
-                connection = options[:connection] || Excon.new(STS_GLOBAL_ENDPOINT, :query => params)
+
+                if ENV["AWS_STS_REGIONAL_ENDPOINTS"] == "regional" && ENV["AWS_DEFAULT_REGION"]
```
geemus, line 46: Could we simplify this by using a single ENV var, maybe `STS_ENDPOINT`and allow this to override the global value? I feel like that may be easier to understand and use than having a regional on/off value and a distinct region value. It's certainly possible I'm missing something here though. Are there precedents on key names I should be aware of?

tests/credentials_tests.rb
MichaelGerhart: Yep, makes sense to me. I just removed "GLOBAL" from the original variable, I can down case this.

tests/credentials_tests.rb
MichaelGerhart: Good catch, I will get this oversight fixed.

lib/fog/aws/credential_fetcher.rb
MichaelGerhart: This is where I had the biggest back and forth personally.

I created this logic based on the environment variable method of configuring this as listed in the AWS SDK documentation here: https://docs.aws.amazon.com/sdkref/latest/guide/setting-global-sts_regional_endpoints.html.
As documented, there are two main ways to define whether or not regional endpoints should be used.

I suppose there are two different ways to go about this:

1) using an sts_regional_endpoints option in the config file - This could leverage the region configuration that *should* already be there. regional + region = regional STS URL. if sts_regional_endpoints is not defined or is not a valid value then default to the global endpoint.

2) The method that I originally took a stab at, which is the AWS supported method for determining this config with environment variables.

I guess a third option could be both, with the config file taking precedence. but the more that I think about it, maybe it makes more sense to support this configuration through the config file rather depending on environment variables? Let me know your thoughts @geemus

geemus: Ah, yeah I suspected there might be a precedent like that you were referring back to.

I think we have some flexibility here to decide. I think we almost always try to support credential files and/or passed in settings first as it's arguably a more universal use case. Then, we usually add ENV as well if it's something that needs to be easily overridden at runtime. So I think we should definitely support config files and passed config, and ENV would be a nice bonus (from my understanding of things). Then again, maybe this would always only be a runtime override? How does your expected usage stack up against these possibilities?

dewnull:
@geemus this is still an issue.

openfly:
this is still an issue and should not have been closed.

geemus:
Thanks for letting me know. It was closed automatically due to inactivity. If anyone has interest/capacity to address the outstanding change requests I would certainly welcome the help.

stanhu:
I cleaned this up in https://github.com/fog/fog-aws/pull/620.

geemus:
Finished up in #620, thanks @stanhu!

PR #610, opened 2021-05-25
Implement AWS TagSpecifications (closes #603)

### LINKED ISSUE #603: Pass tags when creating EBS ?
Hello,

It seems there's no such feature available according to:
https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/compute/run_instances.rb#L24

Would you be kind enough to implement that ?

Best regards, Adam.

eLvErDe:


lib/fog/aws/models/compute/server.rb
```diff
@@ -196,6 +198,48 @@ def run_instance_options
           else
             options.delete('SubnetId')
           end
+          if tag_specifications
```
geemus, line 201: I think it might be good to move this processing down into run_instances directly, as this would allow it to be used from either the requests directly or via the model such as this. I think there are also some helper functions for the formatting the might help you, and that you can see a similar formatting challenge with network information here: https://github.com/fog/fog-aws/blob/5b25c70ba14f8243930f6ea4f26599f13cd73162/lib/fog/aws/requests/compute/run_instances.rb#L134

geemus (changes requested):
Overall this looks good, and I'm glad you got it working. I think we could refine this a bit by moving it to the request and using existing helpers though, could you take a look and see if that makes sense? We should probably also see about adding some testing to help ensure it does what we want and doesn't regress. Does that all make sense? Thanks!

lib/fog/aws/models/compute/server.rb
eLvErDe: Better now ? ;-)

It's still working for me.

geemus: Yes, thank you, hopefully it was pretty straightforward to move across.

geemus (approved):

geemus:
@eLvErDe Thanks again for working through this with me, glad to get things fixed for you and also integrate the fix for others!

PR #611, opened 2021-06-17
fix storage for ruby 3.0

vincentjoseph:
adapt storage#get_object with a block given.  
ruby 3 raises an error when instantiating a proc without a block

@see https://blog.saeloun.com/2019/09/02/ruby-2-7-proc-without-block-warning.html


geemus (approved):
Thanks for the fix!

PR #612, opened 2021-07-07
Do not use multipart copy for files smaller than 5GB

slonopotamus:
multipart_chunk_size is supposed to only set chunk size for multipart uploads.
It should not influence decision whether multipart upload will be used at all.

This change is supposed to speedup copying of files smaller than 5GB because they can be copied by a single efficient request with x-amz-copy-source

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -145,9 +145,9 @@ module Fog
           requires :directory, :key
 
           # With a single PUT operation you can upload objects up to 5 GB in size. Automatically set MP for larger objects.
-          self.multipart_chunk_size = MIN_MULTIPART_CHUNK_SIZE * 2 if !multipart_chunk_size && self.content_length.to_i > MAX_SINGLE_PUT_SIZE
+          if content_length.to_i > MAX_SINGLE_PUT_SIZE
+            self.multipart_chunk_size = MIN_MULTIPART_CHUNK_SIZE * 2 unless multipart_chunk_size
 
-          if multipart_chunk_size && self.content_length.to_i >= multipart_chunk_size
             upload_part_options = options.select { |key, _| ALLOWED_UPLOAD_PART_OPTIONS.include?(key.to_sym) }
             upload_part_options = upload_part_options.merge({ 'x-amz-copy-source' => "#{directory.key}/#{key}" })
             multipart_copy(options, upload_part_options, target_directory_key, target_file_key)


stanhu:
As mentioned in https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html:

> In general, when your object size reaches 100 MB, you should consider using multipart uploads instead of uploading the object in a single operation.

geemus (changes requested):
I'm not quite sure I follow your concern. In the original implementation one can manually set a multipart_chunk_size OR one will be set if the content_length is greater than 5gb. If you one of these things is true AND the content_length is greater than this chunk size it will use multipart upload. Otherwise it will use the normal copy operation. I think that lines up with the desired behavior you describe, so maybe I'm missing something?

Also, trying to think through the implementation specifics, there are a couple of behavior changes that I think we should consider carefully:

In this new version, if `content_length` is greater than `MAX_SINGLE_PUT_SIZE` it will set the `multipart_chunk_size` (if it isn't already set) and then use this to upload things.

In the old version, if `multipart_chunk_size` isn't already set, it will be set if `content_length` is greater than `MAX_SINGLE_PUT_SIZE`. Then, if it is set and the `content_length` is also greater than the multipart_chunk_size it will use multipart uploads.

I believe this creates some gaps in what will actually occur:

1. In the old version, you could opt-in to multipart by directly setting the chunk size, regardless of the content_length. (I'm not sure how important or desirable this is, but it wouldn't be possible in the new version).
2. In the old version, if will not use multipart copy if `multipart_chunk_size` is set but the content_length is less than this size. In the new version it will always be used if the `content_length` is greater than the `MAX_SINGLE_PUT_SIZE`. (again, I'm not sure how important or desirable this is, but it is another behavior change).

Could you let me know what I might be missing about the difference between the current behavior and what you desire, as well as help me understand these behavior changes if we stick to this approach I would appreciate it.

Thanks!

slonopotamus:
> In general, when your object size reaches 100 MB, you should consider using multipart uploads instead of uploading the object in a single operation.

We're not uploading anything, the object is already in the storage, we're just telling the storage "save what you already have under a different name". This is a very important difference.

----

After some more thinking, I possibly want to change this code in a completely different way.

Input data:

1. Not every S3-compatible storage has 5GB threshold at all. Ceph doesn't. So, it should be possible for a caller to say "never use multipart". When storage knows that you want to copy the whole file, it can do that in a much more efficient way, possibly even independent of object size. And we do observe that single-request copy is always faster on Ceph.
2. When using Amazon S3, caller normally shouldn't know the internals of copying process. They just want the file to be copied, in the most efficient way possible. I _believe_ that if we're allowed to use chunks up to size of 5GB, we should use 5GB chunks. I didn't do benchmarks though.
3. Note that current multipart default is 10MB. And it was increased from 5MB because that improved the speed. But I doubt there's something magical about 10MB and using even bigger chunks will be even faster.

geemus:
Thanks, the extra context helps a lot. I'm fairly certain it was written focusing on AWS S3 in particular, which probably explains some of the discrepancy to Ceph you mention. I'll reply to your notes in turn.
1. Having some kind of opt-out, for Ceph or other non-AWS things makes sense. Do you think it should be at the copy-command level, or should it be something you set at a higher level which applies to all copy commands (ie maybe an option on the service)? Having it on the command may be simpler/more direct, but having to set it on every command seems like it could be a nuisance in usage.
2. Agreed that ideally it just does whatever is best, I think that is the intention of the current implementation, though perhaps there is room for improvement. I'm certainly open to tweaking the chunk size, though I would like us to do some benchmarking to be more certain.
3. Good callout. Yes, that is probably somewhat arbitrary, but I'm not certain (it's been ages, so maybe there was a particular reason).
Thanks for the details and continuing to talk through this, hopefully we can find a good balance of things to improve this.

slonopotamus:
I'm gonna open a new PR (with a different solution) and add more context there.

geemus:
Makes sense, those both sound like great goals given our discussion so far. FWIW, it might make sense to split that into two workstreams/PRs to keep things simple and focused.

slonopotamus:
See #616

PR #613, opened 2021-07-10
Update file.rb

iqre8:
Add website_redirect_location

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -268,7 +268,7 @@ module Fog
           options.merge!(metadata)
           options['x-amz-storage-class'] = storage_class if storage_class
           options['x-amz-tagging'] = tags if tags
-          options['x-amz-website-redirect-location' = website_redirect_location if website_redirect_location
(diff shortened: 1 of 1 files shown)


lib/fog/aws/models/storage/file.rb
```diff
@@ -266,6 +268,7 @@ def save(options = {})
           options.merge!(metadata)
           options['x-amz-storage-class'] = storage_class if storage_class
           options['x-amz-tagging'] = tags if tags
+          options['x-amz-website-redirect-location' = website_redirect_location if website_redirect_location
```
geemus, line 271: It looks like, from the tests, that this is missing the closing square bracket `]` just prior to the `=`. Could you please fix that up? Otherwise, looks good.

geemus (changes requested):

iqre8 (commented):
Added the missing ]

lib/fog/aws/models/storage/file.rb
iqre8: Hi @geemus 
Thanks. Updated the file.

geemus (approved):
Thanks for the quick fix!

geemus:
Thanks!

PR #616, opened 2021-07-26
Add storage option to configure multipart put/copy

slonopotamus:
Amazon S3 limits max chunk size that can be uploaded/copied in a single request to 5GB.
Other S3-compatible storages (like, Ceph) do not have such limit.
Ceph shows much better performance when file is copied as a whole, in a single request.

This commit adds `max_put_chunk_size`/`max_copy_chunk_size`storage options
that allow user to configure respective chunk sizes.
Usage of non-positive value will tell fog-aws to use a single put/copy request regardless of file size.

This commit preserves backward compatibility:
if user explicitly sets `multipart_chunk_size` on a `Fog::AWS::Storage::File` instance,
that value will take precedence.

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -14,6 +14,9 @@ module Fog
         'https' => 443
       }
 
+      MIN_MULTIPART_CHUNK_SIZE = 5242880
+      MAX_SINGLE_PUT_SIZE = 5368709120
+
       VALID_QUERY_KEYS = %w[
         acl
         cors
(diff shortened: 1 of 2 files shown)


lib/fog/aws/storage.rb
```diff
@@ -284,15 +314,15 @@ def request_params(params)
               path_style = params.fetch(:path_style, @path_style)
               if !path_style
                 if COMPLIANT_BUCKET_NAMES !~ bucket_name
-                  Fog::Logger.warning("fog: the specified s3 bucket name(#{bucket_name}) is not a valid dns name, which will negatively impact performance.  For details see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html")
+                  Fog::Logger.warning("fog: the specified s3 bucket description(#{bucket_name}) is not a valid dns description, which will negatively impact performance.  For details see: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html")
```
geemus, line 317: The error changes here from using "name" to "description" seem unrelated to other changes, and I worry they might be somewhat confusing/surprising. Could you help me understand them and/or drop them for now to keep this focused? Thanks!

geemus (changes requested):
Looks really good overall, thanks for talking through that with me and adapting your change. This looks really close, and I just had one question around the changes to error messages, I think otherwise it appears ready. Thanks!

lib/fog/aws/storage.rb
slonopotamus: Whoops, I'm notsure how this line got changed. Will revert it.

slonopotamus: Fixed

geemus:
No problem, thanks for fixing it up.

geemus (approved):

slonopotamus:
That would be awesome

geemus:
Released in v3.11.0

PR #619, opened 2021-08-18
Add IPv6 support for Ingress Security Groups

p8:
This change adds adding IPv6 security group rules.

The current security groups tests are pending, so I didn't add any new tests.
This fork does allow me to authorize and revoke IPv6 rules for ingress.
I could not test the egress implementation, as I'm currently not using it.

### DIFF
--- a/lib/fog/aws/models/compute/security_group.rb
+++ b/lib/fog/aws/models/compute/security_group.rb
@@ -62,7 +62,8 @@ module Fog
         # options::
         #   A hash that can contain any of the following keys:
         #    :cidr_ip (defaults to "0.0.0.0/0")
-        #    :group - ("account:group_name" or "account:group_id"), cannot be used with :cidr_ip
+        #    :cidr_ipv6 cannot be used with :cidr_ip
+        #    :group - ("account:group_name" or "account:group_id"), cannot be used with :cidr_ip or :cidr_ipv6
         #    :ip_protocol (defaults to "tcp")
         #
         # == Returns:
@@ -178,7 +179,8 @@ module Fog
         # options::
         #   A hash that can contain any of the following keys:
         #    :cidr_ip (defaults to "0.0.0.0/0")
-        #    :group - ("account:group_name" or "account:group_id"), cannot be used with :cidr_ip
+        #    :cidr_ipv6 cannot be used with :cidr_ip
+        #    :group - ("account:group_name" or "account:group_id"), cannot be used with :cidr_ip or :cidr_ipv6
         #    :ip_protocol (defaults to "tcp")
         #
         # == Returns:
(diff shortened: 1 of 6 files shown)


lib/fog/aws/models/compute/security_group.rb
```diff
@@ -327,9 +327,15 @@ def fetch_ip_permission(range, options)
           }
 
           if options[:group].nil?
-            ip_permission['IpRanges'] = [
-              { 'CidrIp' => options[:cidr_ip] || '0.0.0.0/0' }
-            ]
+            if options[:cidr_ipv6].nil?
```
geemus, line 330: Are the `CidrIp` and `CidrIpv6` mutually exclusive? This implementation appears to reflect that, so I wanted to check-in in case it was possible to do dual stack type stuff. Does that make sense?

geemus (changes requested):
Looks good overall, just want to confirm about if ip and v6 stuff are mutually exclusive.

lib/fog/aws/models/compute/security_group.rb
p8: I could not create a permission with both `CidrIp` and `CidrIpv6`.
With a request body like:
```
Action=AuthorizeSecurityGroupIngress&GroupId=sg-xxxxxxx&IpPermissions.1.FromPort=443&IpPermissions.1.IpProtocol=tcp&IpPermissions.1.IpRanges.1.CidrIp=1.2.3.4%2F0&IpPermissions.1.Ipv6Ranges.1.CidrIpv6=xxxx%3xxxx%3Axxxx%3Axxxx%3A%3A%2F64&IpPermissions.1.ToPort=443&Version=2016-11-15
```

It would return an error like:
```xml
<Code>InvalidPermission.Duplicate</Code><Message>the specified rule \\"peer: 0.0.0.0/0, TCP, from port: 443, to port: 443, ALLOW\\" already exists</Message></Error>
```
It is also not supported in the AWS web console.

geemus: Thanks for confirming, I suspected something like that might be the case, but wanted to double check in case we might be missing something.

geemus (approved):

p8:
Thanks @geemus !

geemus:
@p8 Thanks for the help! Do you need a release with these changes, or is it ok to wait a bit?

It would be great to get the existing tests out of pending, should you have interest/capacity at some point.

p8:
@geemus I'm in no hurry 🙂 .
I'll see if I can get the tests to run.

geemus:
@p8 appreciated, just let me know if you have questions.

P.S. Changes should be available in v3.12.0

PR #620, opened 2021-08-20
 Add support for regional STS endpoints

stanhu:
This pull request is based off https://github.com/fog/fog-aws/pull/605/files and cleans up and fixes the implementation and test cases.

### DIFF
--- a/lib/fog/aws/credential_fetcher.rb
+++ b/lib/fog/aws/credential_fetcher.rb
@@ -21,7 +21,7 @@ module Fog
           if options[:use_iam_profile]
             begin
               role_data = nil
-              region = options[:region]
+              region = options[:region] || ENV["AWS_DEFAULT_REGION"]
 
               if ENV["AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"]
                 connection = options[:connection] || Excon.new(CONTAINER_CREDENTIALS_HOST)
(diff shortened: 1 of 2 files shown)


stanhu:
@geemus Could you take a look at this?

lib/fog/aws/credential_fetcher.rb
```diff
@@ -23,7 +21,7 @@ def fetch_credentials(options)
           if options[:use_iam_profile]
             begin
               role_data = nil
-              region = options[:region]
+              region = options[:region] || ENV["AWS_DEFAULT_REGION"]
```
stanhu, line 24: I think `AWS_REGION` might have been used before, but I think `AWS_DEFAULT_REGION` is the new environment variable: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html.

lib/fog/aws/credential_fetcher.rb
```diff
@@ -44,7 +42,15 @@ def fetch_credentials(options)
                   :WebIdentityToken => File.read(options[:aws_web_identity_token_file] || ENV.fetch("AWS_WEB_IDENTITY_TOKEN_FILE")),
                   :Version => "2011-06-15",
                 }
-                connection = options[:connection] || Excon.new(STS_GLOBAL_ENDPOINT, :query => params)
+
+                sts_endpoint =
+                  if ENV["AWS_STS_REGIONAL_ENDPOINTS"] == "regional"
```
stanhu, line 47: Maybe we need to check `region` as well.

lib/fog/aws/credential_fetcher.rb
```diff
@@ -44,7 +42,15 @@ def fetch_credentials(options)
                   :WebIdentityToken => File.read(options[:aws_web_identity_token_file] || ENV.fetch("AWS_WEB_IDENTITY_TOKEN_FILE")),
                   :Version => "2011-06-15",
                 }
-                connection = options[:connection] || Excon.new(STS_GLOBAL_ENDPOINT, :query => params)
+
+                sts_endpoint =
+                  if ENV["AWS_STS_REGIONAL_ENDPOINTS"] == "regional" && region
```
geemus, line 47: Part of this still seems a little weird to me. I guess it feels like maybe the key should just be something like `true`? ie `ENV["AWS_STS_REGIONAL_ENDPOINTS"]` is binary, so making it that explicitly seems a bit cleaner to me than using a special `regional` key. Does that make sense?

geemus: Ah, wait, I see. It matches the keys/values from the AWS CLI. That makes more sense.

geemus (approved):
LGTM, thanks for helping finish this up!

stanhu:
@geemus Thanks for the quick review! Would you mind tagging a new release?

MichaelGerhart:
@stanhu thanks for getting this finished up! I meant to come back to it but there’s always something…

geemus:
@MichaelGerhart no worries, we've all been there.

geemus:
@stanhu is Monday soon enough? I'm always wary of releasing right before the weekend (when I won't be checking email and stuff as much).

stanhu:
Sure.

geemus:
Awesome, thanks!

geemus:
Released in v3.12.0.

PR #626, opened 2021-10-29
Update Fog::AWS::Storage::File#body

10io:
Port of https://github.com/fog/fog-google/pull/545 where:
- last_modified condition was removed
- method has been rewritten to be simpler


geemus (approved):
LGTM, thanks!

PR #628, opened 2021-11-13
Add Truffleruby head to CI

gogainda:


geemus:
@gogainda I don't have that much context around truffleruby, particularly head. How stable is it and is this likely to lead to broken builds? Should we add it to allowed failures? Thanks.

gogainda:
it should be as stable as ruby head, but I dont mind having it with experimental tag

geemus (approved):

geemus:
K, thanks. Will bring it in for now and we can always add to experimental if we see issues.

PR #629, opened 2021-11-18
Exclude test files from gem

ursm:
This gem contains test files totaling 1.4MB, which can be excluded from the package since they should not be needed at runtime.

```
$ du -sh *
80K	CHANGELOG.md
4.0K	CONTRIBUTING.md
8.0K	CONTRIBUTORS.md
4.0K	LICENSE.md
4.0K	README.md
4.0K	fog-aws.gemspec
6.1M	lib
1.4M	tests
```

According to https://github.com/rubygems/guides/issues/90, `spec.test_files` is no longer in use.


geemus (approved):
Looks good, thanks for the details and fix!

PR #637, opened 2022-03-05
Add option to control IAM credential refresh

hchouraria:
Prior to this change the behavior of the library has been
to refresh the credential from IAM only after its in the last
15 seconds of its life according to its expired_at field

This poses a problem when using the library to generate short
lived pre-signed URLs in AWS S3 where the issued URLs
with temporary tokens expire the same time as the IAM credential
that generated them:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html

For example, if the credentials is in its last 16 seconds of life
and a token has been generated for pre-signed URLs with a life
request of 10 minutes, the URL ends up living only 16 seconds.

By allowing the caller to configure the refresh threshold a
larger buffer of time can be specified when using IAM credentials
allowing for the generated URLs to life a more definitive time.

This change adds a configuration option that can be passed by
callers to expand (or narrow) the threshold of time used by the
refresh credentials mechanism to decide if a new token needs
to be fetched.

Configuration option is called
'aws_credentials_refresh_threshold_seconds'

Following the same example, setting this new option to 10 minutes
can help guarantee that pre-signed URLs generated for 10 minutes
will always last that long.

A real world example of the narrow default of 15 seconds posing a problem can be read on GitLab's issue tracker: https://gitlab.com/gitlab-org/gitlab/-/issues/350118 (GitLab's Pages sub-component caches generated pre-signed URLs up to 10 minutes for performance reasons)


hchouraria:
While the source issue that required this was specific to AWS S3, I wasn't sure if other AWS products have similar temporary tokens that only last as long as the IAM credential so I've added the configurable support to all applicable classes that had a `setup_credentials` phase in it. If this is excessive/unnecessary I can reduce it to only touch `storage.rb`.

geemus (commented):
Hey, thanks for the detailed report. Overall I think this looks good and makes sense, and I had a couple points I wanted to discuss.

1. I don't think there are other timed/expiry things in the other services, but it's certainly possible I'm simply forgetting (I haven't been using fog directly myself in some time). So it might be better to limit this to S3, where we know it is needed, to avoid additional complexity that we might not need until we know it is needed?
2. You did a great job describing the particular edge case you were hitting and how this addresses it in the issue, and I'm worried that we aren't transferring the context into the documentation and/or code for posterity. Does that make sense? I think otherwise others may overlook and/or not understand the value of this.

What do you think? Thanks!

hchouraria:
Thank you for taking a look @geemus, I've addressed the comments in a new push, please let me know if further changes are required!

geemus (approved):
Looks good, thanks for working through it with me!

PR #641, opened 2022-04-30
Create codeql.yml

naveensrinivasan:
This action runs GitHub's industry-leading semantic code analysis engine, CodeQL, against a repository's source code to find security vulnerabilities.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql

https://github.com/ossf/scorecard/blob/main/docs/checks.md#sast

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,67 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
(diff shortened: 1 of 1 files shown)


geemus (approved):

geemus (changes requested):
Looks good. Could you help us fix the issues this revealed as well? It looks like it should be just adding escapes to the dots in those urls, so nothing fancy. Now that I know about them, I guess I'd prefer to take care of them now before we forget or they get lost. Thanks!

naveensrinivasan:
> Looks good. Could you help us fix the issues this revealed as well? It looks like it should be just adding escapes to the dots in those urls, so nothing fancy. 

I don't know ruby and don't want to change things that I can't test.

>Now that I know about them, I guess I'd prefer to take care of them now before we forget or they get lost. Thanks!

They won't get lost. It will be there in the security tab. 

I hope you understand. Thanks

geemus:
Yes, I think so. Thanks.

geemus (approved):

PR #643, opened 2022-05-04
RDD tags issue in AWS GovCloud Account regions.

svavhal:
Fixed RDS tags issue for AWS GovCloud Account regions.

### DIFF
--- a/lib/fog/aws/requests/rds/add_tags_to_resource.rb
+++ b/lib/fog/aws/requests/rds/add_tags_to_resource.rb
@@ -12,28 +12,43 @@ module Fog
         #   * body<~Hash>:
         def add_tags_to_resource(rds_id, tags)
           keys    = tags.keys.sort
-          values  = keys.map {|key| tags[key]}
+          values  = keys.map { |key| tags[key] }
+          resource_name = "arn:aws:rds:#{@region}:#{owner_id}:db:#{rds_id}"
+          %w[us-gov-west-1 us-gov-east-1].include?(@region) ? resource_name.insert(7, '-us-gov') : resource_name
           request({
-              'Action'        => 'AddTagsToResource',
-              'ResourceName'  => "arn:aws:rds:#{@region}:#{owner_id}:db:#{rds_id}",
-              :parser         => Fog::Parsers::AWS::RDS::Base.new,
-            }.merge(Fog::AWS.indexed_param('Tags.member.%d.Key', keys)).
-              merge(Fog::AWS.indexed_param('Tags.member.%d.Value', values)))
+            'Action' => 'AddTagsToResource',
+            'ResourceName' => resource_name,
+            :parser => Fog::Parsers::AWS::RDS::Base.new
+          }.merge(Fog::AWS.indexed_param('Tags.member.%d.Key', keys))
+              .merge(Fog::AWS.indexed_param('Tags.member.%d.Value', values)))
+        end
+
+        def add_tags_to_dbsnapshot(snapshot_name, tags)
+          keys    = tags.keys.sort
+          values  = keys.map { |key| tags[key] }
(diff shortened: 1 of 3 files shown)


lib/fog/aws/requests/rds/add_tags_to_resource.rb
```diff
@@ -12,28 +12,43 @@ class Real
...
-            }.merge(Fog::AWS.indexed_param('Tags.member.%d.Key', keys)).
-              merge(Fog::AWS.indexed_param('Tags.member.%d.Value', values)))
+            'Action' => 'AddTagsToResource',
+            'ResourceName' => resource_name,
+            :parser => Fog::Parsers::AWS::RDS::Base.new
+          }.merge(Fog::AWS.indexed_param('Tags.member.%d.Key', keys))
+              .merge(Fog::AWS.indexed_param('Tags.member.%d.Value', values)))
+        end
+
+        def add_tags_to_dbsnapshot(snapshot_name, tags)
```
geemus, line 26: For consistency I think we might want this to be defined in it's own request file? Presumably `lib/fog/aws/requests/rds/add_tags_to_dbsnapshot.rb`. That will make it easier to find in the future, should we need it. It would also be good to see some tests for this and probably also so Mocks to round it out and provide the same level of integration and functionality as the other functions. Does that make sense? What do you think?

lib/fog/aws/requests/rds/list_tags_for_resource.rb
```diff
@@ -11,27 +11,40 @@ class Real
...
-            'Action'        => 'ListTagsForResource',
-            'ResourceName'  => "arn:aws:rds:#{@region}:#{owner_id}:db:#{rds_id}",
-            :parser         => Fog::Parsers::AWS::RDS::TagListParser.new
+            'Action' => 'ListTagsForResource',
+            'ResourceName' => resource_name,
+            :parser => Fog::Parsers::AWS::RDS::TagListParser.new
           )
         end
+
+        def list_tags_for_snapshot(snapshot_name)
```
geemus, line 25: Similarly here, I think we might want to pull this out to something like: `lib/fog/aws/requests/rds/list_tags_for_snapshot.rb` and add tests and mock functionality.

geemus (changes requested):
Thanks for the fixes, I think overall they look good. I think the resource fixes look like they are probably more or less done, but I had some questions about the snapshot methods. Those seem like feature additions at this point, rather than a bug fix, so it might make things cleaner/smoother to separate them out into a separate pull request. That could also allow us to get these bug fixes in promptly, while  allowing time to work through getting the additions polished up and added. Does that make sense? Are you up for separating the two things so we can address them independently? Thanks!

lib/fog/aws/requests/rds/list_tags_for_resource.rb
svavhal: @geemus Yes
 removed rds snapshot changes will add in another PR.

lib/fog/aws/requests/rds/add_tags_to_resource.rb
svavhal: yes right @geemus 
removed rds snapshot changes from this PR
Will in another PR.

geemus (approved):
Thanks for understanding, this looks good, thanks!

geemus:
Did you need a release with this fix?

svavhal:
> Thanks for understanding, this looks good, thanks!
thanks @geemus 


> Did you need a release with this fix?
Yes please

geemus:
@svavhal Sounds good. Thanks again for the fix. I'll plan to release early next week (I'm not online over weekends as much, so I prefer to release when I'm more likely to see if there are problems and can respond quickly).

geemus:
Released as part of v3.14.0. Thanks again!

PR #646, opened 2022-05-10
Fix a typo in CHANGELOG

y-yagi:
It seems that `RDD` is a typo of `RDS`.


geemus:
Thanks. It's generated from the pull requests (which also had the typo), so I fixed the title of that as well (lest we overwrite this next time we regenerate things).

geemus (approved):

PR #649, opened 2022-06-24
Update README file with download url example

lucasocon:
Add an example to get a download URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,20 @@ directory.files
 directory.files.new(key: 'user/1/Gemfile').url(Time.now + 60)
 ```
 
+##### Generate download URL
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,6 +106,20 @@ directory.files
 directory.files.new(key: 'user/1/Gemfile').url(Time.now + 60)
 ```
 
+##### Generate download URL
+You should pass an option arguyment that contains the `query` key with `response-content-disposition` inside indicating that is an attachment and the filename to be used when downloaded.
```
geemus, line 110: It looks like there is a typo here `arguyment` that should be `argument`.

geemus (changes requested):
Thanks, looks good overall. I just saw one small typo. If you could fix that we should be able to get this in soon. Thanks!

geemus (approved):
I'll just go ahead and pull this in and then make the fix myself since this seems to have gotten lost.

PR #651, opened 2022-07-20
change sync_clock to plain GET

duckworth:
Perform an unauthenticated GET request to the root S3 URL to prevent 403 errors if the user doesn't have ListBuckets permissions


geemus (approved):

geemus:
@duckworth Looks good to me, thanks for talking through this and following up with the patch.

Issue #653, opened 2022-09-04
URI.decode is obsolete (and not available in Ruby 3.0)


kuahyeow:
See https://ruby-doc.org/stdlib-2.5.3/libdoc/uri/rdoc/URI/Escape.html#method-i-unescape

It seems there's 4 usages still - https://github.com/fog/fog-aws/search?q=URI.decode

geemus:
Thanks for the heads up. One possibility might be to use the solution we used for this in Excon, or something similar. As I recall there are quite a few caveats and potential edge cases in whatever solution one might choose though.

Here is the excon solution: https://github.com/excon/excon/blob/58a8d9511908c9cd8aa7e3649106eb7e3740cc8e/lib/excon/utils.rb#L103

Do you think something like that would be sufficient for our purposes? And/or do you have ideas of alternatives?

geemus:
Sorry for the lack of context. Excon is an http library that fog uses for it's api calls, and that was originally embedded in fog before being extracted as it's own thing. So, under the covers fog is using it for a lot of this stuff and I've worked on both a lot, which is why that example came to mind.

I can't presently recall why we did it directly in Excon instead of using `URI.decode_www_form_component`. I think in an effort to maintain broader backwards compatibility.

All that said, I think you are probably right that in this case we could keep things simple and address the issue by using that method. Would you be up for making a pull request with that fix for us? Thanks!

kuahyeow:
Oh I see, it looks like https://github.com/excon/excon/issues/334. so `decode_www_form_component` appears to be available since Ruby 1.9, and this gem supports Ruby >= 2.0.0 so we should be OK.

Sure,  I will make a pull request.

kuahyeow:
I have opened https://github.com/fog/fog-aws/pull/654 now

geemus:
@kuahyeow Good find, it somehow didn't occur to me to search back through the PRs to find that. Definitely makes sense that it would be an old version issue and also that it should be ok here as fog is more restrictive on versions than excon is/was. Thanks for checking in to that and refreshing my memory. I'll look at the PR presently.

PR #654, opened 2022-09-07
Replace URI.decode as obsolete, and removed in Ruby 3.0

kuahyeow:
Resolves https://github.com/fog/fog-aws/issues/653

Replace with URI.decode_www_form_component which has been available since Ruby 1.9

I'm not sure if there's any tests to add / update.

Based on https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/lint/uri_escape_unescape.rb, we have three options:

1. CGI.escape
2. URI.encode_www_form
3. URI.encode_www_form_component

2 is complicated because it returns an Array instead of a String. 3 seems a better fit than 1. Also:

```
irb(main):001:0> string = '?a=%09%0D'
=> "?a=%09%0D"
irb(main):002:0> URI.decode_www_form_component(string) == URI.decode(string)
(irb):2: warning: URI.unescape is obsolete
=> true

irb(main):004:0> require 'cgi'
=> true
irb(main):005:0> CGI.escape(string)
=> "%3Fa%3D%2509%250D"
irb(main):006:0> CGI.escape(string) == URI.decode(string)
(irb):6: warning: URI.unescape is obsolete
=> false
```

### DIFF
--- a/lib/fog/aws/parsers/iam/get_group_policy.rb
+++ b/lib/fog/aws/parsers/iam/get_group_policy.rb
@@ -14,7 +14,7 @@ module Fog
             when 'GroupName', 'PolicyName'
               @response[name] = value
             when 'PolicyDocument'
-              @response['Policy'][name] = if decoded_string = URI.decode(value)
+              @response['Policy'][name] = if decoded_string = URI.decode_www_form_component(value)
                                   Fog::JSON.decode(decoded_string) rescue value
                                 else
                                   value
--- a/lib/fog/aws/parsers/iam/get_role_policy.rb
+++ b/lib/fog/aws/parsers/iam/get_role_policy.rb
@@ -12,7 +12,7 @@ module Fog
             when 'RoleName', 'PolicyName'
               @response['Policy'][name] = value
             when 'PolicyDocument'
-              @response['Policy'][name] = if decoded_string = URI.decode(value)
+              @response['Policy'][name] = if decoded_string = URI.decode_www_form_component(value)
                                   Fog::JSON.decode(decoded_string) rescue value
                                 else
                                   value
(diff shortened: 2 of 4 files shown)


geemus (approved):
LGTM and thanks for the detailed write up and digging into some of the history from excon to help me remember all the context.

geemus:
@kuahyeow Thanks again, does this need a release as well? Otherwise we tend to do them relatively regularly, but not on every PR.

kuahyeow:
Awesome, thanks for the review, and merge 👍 

As for release, we are no particular rush as we are still on Ruby 2.7. The next regular release sounds good.

mvz:
Unfortunately, Ruby 2.7 will allow updating the default url gem, triggering this issue.

geemus:
@mvz I'm not familiar with the change you mention, could you share a bit more detail about it and the impact we might expect?

mvz:
@geemus sorry, I meant the 'uri' gem, not the 'url' gem.

geemus:
@mvz Ah, yes. I was aware of that. We can certainly consider having our own implementation or otherwise changing around how we do this if and when that becomes an issue.

mvz:
@geemus I think we're talking past each other.

What I'm saying is that the issue fixed by this pull request can already occur on Ruby 2.7, since I can run `gem update uri` on Ruby 2.7 right now and it will trigger the issue. This was in response to @kuahyeow saying that he does not require a release since he's still on Ruby 2.7.

The fix itself is fine afaics.

geemus:
@mvz Ah, ok. Thank you for clarifying. I never like releasing right before the weekend (just in case), but should be able to release early next week.

geemus:
Now released in v3.15.0. Just let me know if anyone runs into any more related issues and we can keep working on it.

PR #669, opened 2023-02-09
Only compute SSE-C headers when needed in multipart upload

stanhu:
Previously fog-aws would attempt to generate all the SSE-C headers even if encryption were not active. However, on a FIPS system, MD5 is not available. This commit modifies the UploadPart API request to compute the SSE-C headers only when they are needed.

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -389,9 +389,16 @@ module Fog
 
         def part_headers(chunk, options)
           base_headers = part_checksum_headers(chunk)
-          encryption_keys = encryption_customer_key_headers.keys
-          encryption_headers = options.select { |key| encryption_keys.include?(key) }
-          base_headers.merge(encryption_headers)
+
+          # Only SSE-C headers needed in the UploadPart request. [1]
+          # x-amz-server-side-encryption and
+          # x-amz-server-side-encryption-aws-kms-key-id are only needed
+          # in the CreateMultipartUpload request. [2]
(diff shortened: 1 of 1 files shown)


lib/fog/aws/models/storage/file.rb
```diff
@@ -387,11 +387,18 @@ def encryption_headers
...
-          encryption_headers = options.select { |key| encryption_keys.include?(key) }
-          base_headers.merge(encryption_headers)
+
+          # Only SSE-C headers needed in the UploadPart request. [1]
+          # x-amz-server-side-encryption and
+          # x-amz-server-side-encryption-aws-kms-key-id are only needed
+          # in the CreateMultipartUpload request. [2]
+          # [1] https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html
+          # [2] https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html
+          base_headers.merge!(encryption_customer_key_headers) if encryption && encryption_key
```
stanhu, line 399: Hmm, are `encryption` and `encryption_key` actually available here?

stanhu: Yes, it looks like `Fog::Model` gets initalized with `options` already, so `encryption` and `encryption_key` will be set.

geemus (approved):
LGTM, thanks!

stanhu:
@geemus Thanks! Would you mind tagging a new release?

geemus:
@stanhu released in v3.18.0.

PR #673, opened 2023-04-19
s3: remove hardcoded host

### LINKED ISSUE #672: bug: directories method ignores :endpoint config
when using fog-aws with a S3 compatible provider there are cases when the endpoint setting is ignored
for example get_service.rb has `s3.amazonaws.com` string hard-coded

honzasterba:
fixes #672

### DIFF
--- a/lib/fog/aws/requests/storage/get_service.rb
+++ b/lib/fog/aws/requests/storage/get_service.rb
@@ -21,7 +21,7 @@ module Fog
           request({
             :expects  => 200,
             :headers  => {},
-            :host     => 's3.amazonaws.com',
+            :host     => @host || region_to_host(DEFAULT_REGION),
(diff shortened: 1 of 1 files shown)


geemus:
Thanks. I'm struggling to remember why that change was made in the first place, looks like it was back in 2019. My only concern is if this needs to always go to the main s3 host instead of a region-based host. We might need to be more sophisticated about it (ie any of the aws stuff should maybe map to this value, instead of the regional one). Are you able to test to see what happens here?

honzasterba:
updated to use the default region host unless @host has been set

geemus (approved):

geemus:
Thanks, I think/hope that should cover our bases.

geemus:
Released in v3.19.0.

Issue #674, opened 2023-05-12
S3 upload retrying


rajyan:
We're currently using CarrierWave with fog-aws to upload a large number of files to S3 every day.
However, we sometimes encounter errors such as 500, 503, and Excon::Error::Socket intermittently.
We've found similar issues related to S3 retries on GitHub, such as fog/fog-aws#264 and carrierwaveuploader/carrierwave#2610.

After investigating the issue, we discovered that the fog storage upload request is configured as an idempotent request and is already retrying ~~three~~ four times using Excon, as explained in this [comment](https://github.com/fog/fog-aws/issues/264#issuecomment-228859340). However, ~~further investigation revealed that Excon seems [not retrying on 500 nor 503 errors](https://github.com/excon/excon/blob/1e0c0ee3e975b2470262c6c882dccc752382e03c/lib/excon/constants.rb#L19-L23) (if my understanding is correct)~~ wasn't true. See the conversation below. Additionally, the default `retry_interval` of 0 seconds may be too short.

Therefore, we changed our Excon configuration for fog storage to the following:

```ruby
{
  retry_limit: 5,
  retry_interval: 5,
  retry_errors: [
    Excon::Error::Timeout,
    Excon::Error::Socket,
    Excon::Error::HTTPStatus,
    Excon::Error::InternalServerError,
    Excon::Error::ServiceUnavailable,
  ]
}
```
This change greatly improved our S3 retries, and we haven't encountered an S3 error for a month with this configuration.

Do you think it would be reasonable to change the default configuration for fog storage to this?

geemus:
Thanks for the detailed report.

I think the server errors (ie 500/503) are just a child to HTTPStatus errors, so that they would be retried already, see: https://github.com/excon/excon/blob/1e0c0ee3e975b2470262c6c882dccc752382e03c/lib/excon/error.rb#L83.

Bumping the interval and retries is probably fine, though a nicer solution might be something like exponential backoff (for some reason I kind of thought that is what we already did, but I must be misremembering as I can't find anything presently that looks like that).

Going from 4 to 5 retries seems fine, I guess bumping from 0 to 5 seconds on the retry seems like a much bigger jump. Could we start with something smaller, maybe 1 or 2? As I said above, I kind of feel like backoff might be ideal, but that is more involved in terms of implementation. What do you think?

rajyan:
> I think the server errors (ie 500/503) are just a child to HTTPStatus errors, so that they would be retried already,

[...]

PR #675, opened 2023-05-15
configure connection_options to improve s3 retry

rajyan:
closes https://github.com/fog/fog-aws/issues/674

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -546,7 +546,7 @@ module Fog
           @use_iam_profile = options[:use_iam_profile]
           @instrumentor       = options[:instrumentor]
           @instrumentor_name  = options[:instrumentor_name] || 'fog.aws.storage'
-          @connection_options     = options[:connection_options] || {}
+          @connection_options     = options[:connection_options] || { retry_limit: 5, retry_interval: 1 }
           @persistent = options.fetch(:persistent, false)
           @acceleration = options.fetch(:acceleration, false)
           @signature_version = options.fetch(:aws_signature_version, 4)


geemus:
Thanks!

rajyan:
Thanks for your quick responses!

codez:
I did not look into the details, but the retry_interval of 1 adds 1 second to s3 head requests if the object does not exist. We use fog-aws with carrierwave and check the existence of a file quite often. For example, running some tests with 20 existence checks now take over 20 seconds, when it was about 1 second before. 

Should this be fixed in carrierwave or are there better approaches? We currently configured the retry_interval to 0 again.

geemus:
@codez I think the right thing to do might be pretty context dependent. The retries are in place to account for network inconsistencies and eventual consistency (both of which are pretty common). Setting the retry_interval to 0 should probably be fine if the concern is with network issues, but not as much eventual consistency. That might not be the right balance for everyone though. Also, it may be that one tuning of retry makes sense for your head checks, but you would want a different one for other calls. ie you may want to treat it differently if you believe that the object will exist or not vs objects that may just not exist yet (but will once eventual consistency does it's thing). That being said, you may be ok with eventually consistent things that aren't there yet just returning an error also.

Does that make sense? I'm not sure given the various caveats I outlined what the right general approach is, though I suspect from what you've said that setting the interval to 0 may be sufficient for your purposes.

codez:
Thank you very much for your detailed answer, that makes perfectly sense. I also think it would be best to use a specific retry interval for the head requests. I will ponder if this is possible with carrierwave.

geemus:
@codez Yeah, I think mostly the retries are kind of assumed to be global, rather than on a per request or per request type. I don't recall having discussed these particular needs before, so maybe it just hadn't quite come up like this. You could potentially change those settings before and after doing head requests I suppose, though it could introduce some risks around concurrency or having the state remain if there was an error among other things. I don't recall carrierwave providing this possibility directly either.

codez:
I haven't found a possibility so far. Carrierwave keeps a single Fog::Storage instance for all operations. Changing configs for each request sounds quite hacky ;). As we did fine with a zero retry interval so far, we will just keep it configured like this.

geemus:
@codez Yeah, that sounds right to me. As long as you are aware of the possible ramifications of eventual consistency (ie it might take a while after writing a key before it will start appearing in responses, and even then some responses might succeed while others fail for a little while), I think the 0 retry should probably be fine.

rajyan:
I haven't looked into the details yet, but I think that we shouldn't retry on Client errors?

Maybe we should change `Excon::Error::HTTPStatus` to `Excon::Error::Server` here.
https://github.com/excon/excon/blame/1e0c0ee3e975b2470262c6c882dccc752382e03c/lib/excon/constants.rb#L19-L23

rajyan:
Looking at this comment
https://github.com/excon/excon/issues/689#issuecomment-454534684

I'm now thinking that it might be better to revert this, and configure these settings on the user side (which is carrirerwave in our cases).

rajyan:
Taking a quick look

boto is retrying on general connection errors
https://github.com/boto/botocore/blob/2322be09d78dc51a9187b3a71101532c573b246a/botocore/retryhandler.py#L33-L40

aws-sdk-js is retrying on transient errors and throttling errors.
https://github.com/aws/aws-sdk-js-v3/blob/864b34a52aea8dbd52a63239027b4c203c75863c/packages/util-retry/src/StandardRetryStrategy.ts#L63

geemus:
@rajyan I believe you can pass through options to excon by setting values for `connection_options` in the credentials used, and that you could use this here to override the `retry_errors` key and change or omit things if need be. Could you try that and see it if gets you the behavior you want? That feels a bit safer in terms of potential side effects and impact than reverting the changes mentioned. What do you think?

PR #680, opened 2023-05-23
Change method sync_clock_url to use localstack set a different port

fabiodallazen:
Correct the sync_clock_url method so that it is possible to connect to a port other than the default one.
Use case where it needs to be a different port than the default one, is using localstack s3 in a development environment.


geemus (approved):
LGTM

Issue #683, opened 2023-06-22
When using IRSA on `Fog::Storage`, output `Unrecognized arguments: sts_endpoint` warning 


nekomaho:
Version: v3.19.0

I pass the following config to connect S3 by [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) using `Fog::Storage`.
```ruby
fog_params = {
  provider: 'AWS',
  region: 'ap-northeast-1',
  use_iam_profile: true,
}
ENV['AWS_ROLE_ARN'] = 'xxxx'
ENV['AWS_STS_REGIONAL_ENDPOINTS'] = 'region'

fog=Fog::Storage.new(fog_params)
```

Then output the following warning. 
```
[fog][WARNING] Unrecognized arguments: sts_endpoint
```
But, in spite of the above warning, the connection to S3 has been successful.
So, this warning seems to be only a display problem.
And this error occurs from the following code.
https://github.com/fog/fog-core/blob/v2.3.0/lib/fog/core/service.rb#L250

I have investigated why this warning occurs and found it seems to be `sts_endpoint` is not include `recognizes` values that are causing that warning.
https://github.com/fog/fog-aws/blob/v3.19.0/lib/fog/aws/storage.rb#L49

In addition, I have tried to add `sts_endpoint` to `recognizes` of `AWS::Fog::Storage` `recognizes`; that warning no longer outputs.

I would like to ask you if `Unrecognized arguments: sts_endpoint` is the correct warning or not.
And if correct, please tell me how to avoid this warning if you know.

geemus:
Thanks for the detailed report. As you noticed, I believe the warning is just a warning, so it should work even if/when it warns.

That being said, I think this is just an oversight. I suspect we missed this case for the recognizes list when we later added the sts options. I think the correct fix is likely to add the `:sts_endpoint` key to the list of recognized warnings. Does that match with your understanding? Would you be up for creating the PR with the fix?

Thanks!

nekomaho:
Thank you for your response!

> I think the correct fix is likely to add the :sts_endpoint key to the list of recognized warnings. Does that match with your understanding?

Yes, it's the same as my understanding.
I'll try to create PR with the fix.

geemus:
Sounds good, thanks!

PR #684, opened 2023-06-28
Prevent `Unrecognized arguments: sts_endpoint` warning

nekomaho:
This is a solution for https://github.com/fog/fog-aws/issues/683

## Description

I try to fix three points in this PR.

1. Fix `[fog][WARNING] Unrecognized arguments: sts_endpoint` warring
   - this is the main topic since #683

2. Add another related parameter such as a `region`
   - In coding, I found the `Unrecognized arguments:` warning is not only `sts_endpoint`. So I try to fix other parameters' warnings.

3. Add tests that prevent those warnings from automatically
   - Currently, we need to manually cross-check how to correct parameters that are set on each AWS Service class when the corresponding AWS new services or add extra parameters to CredentialFetcher.
   - So I try to do that operation in the test.

### DIFF
--- a/tests/credentials_tests.rb
+++ b/tests/credentials_tests.rb
@@ -1,6 +1,31 @@
 # frozen_string_literal: true
 
 Shindo.tests('AWS | credentials', ['aws']) do
+  unchecked_constants = [:CredentialFetcher, :VERSION, :Errors, :Mock, :ServiceMapper, :Federation, :STS, :SignatureV4, :EFS]
+
+  test_services = Fog::AWS.constants.delete_if { |service| unchecked_constants.include?(service) }
+
+  target_services = test_services.each_with_object([]) do |name, services|
+    services << Class.const_get("Fog::AWS::#{name}")
+  end
+  require_credential_parameters = [:aws_access_key_id, :aws_secret_access_key]
+  recognized_credential_parameters = [:aws_session_token, :aws_credentials_expire_at, :region, :sts_endpoint]
+
+  # ex: target_services = [Fog::AWS::Compute, Fog:AWS::DNS...]
+  target_services.each do |service|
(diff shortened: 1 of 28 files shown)


tests/credentials_tests.rb
```diff
@@ -1,6 +1,31 @@
 # frozen_string_literal: true
 
 Shindo.tests('AWS | credentials', ['aws']) do
+  unchecked_constants = [:CredentialFetcher, :VERSION, :Errors, :Mock, :ServiceMapper, :Federation, :STS, :SignatureV4, :EFS]
```
nekomaho, line 4: The reason for the above constructs' exclusion from the test is below.

1. CredentialFetcher, VERSION, Errors, Mock, ServiceMapper are not inherited Fog::Service. 

2. Federation, SignatureV4 are not `requires` and `recognizes` defined.

3. STS is `aws_access_key_id` and `aws_secret_access_key` parameters are defined at `recognizes` instead of `requires`.

4. EFS is not defined `aws_session_token` on `recognizes`.

tests/credentials_tests.rb
```diff
@@ -1,6 +1,31 @@
 # frozen_string_literal: true
 
 Shindo.tests('AWS | credentials', ['aws']) do
+  unchecked_constants = [:CredentialFetcher, :VERSION, :Errors, :Mock, :ServiceMapper, :Federation, :STS, :SignatureV4, :EFS, :Parsers]
```
nekomaho, line 4: `Parsers` is not inherited Fog::Service

https://github.com/fog/fog-aws/blob/v3.19.0/lib/fog/aws/parsers/lambda/base.rb#L1-L3

geemus (approved):
LGTM, thanks!

PR #691, opened 2023-09-25
Fog::AWS::Storage don't retry client errors

### LINKED ISSUE #690: Fog::AWS::Storage default retry behaviour guarantees 6s delay for 4xx class responses
The default retry configuration for `Fog::AWS::Storage` https://github.com/fog/fog-aws/blob/b073ba81288af57bcce9144bdec4a77f5c6317b4/lib/fog/aws/storage.rb#L549

combined with Excon's defaults leads to a situation where a GET of an S3 object that may or may not exist will take 6 seconds to answer "no - the object was not found".

That retry configuration was added in https://github.com/fog/fog-aws/issues/674 solving for a batch upload scenario.

This can be reproduced with something like:
```
s3 = Fog::Storage.new(provider: "AWS")
bucket = s3.directories.new(key: 'some-bucket')
puts Benchmark.measure { file = bucket.files.get('some-non-existent-object') }
```
I have a [script that does that for various retry configurations](https://github.com/rahim/fog-s3-issue/blob/main/repro_fog_retry_issue.rb), example results:
```
Fog/Excon defaults (retry_limit: 5, retry_interval: 1)

  0.227765   0.018269   0.246034 (  6.307515)


Using retry_limit: 5, retry_interval: 0.1

  0.242438   0.018477   0.260915 (  2.534994)


Using retry_limit: 1, retry_interval: 1

  0.046131   0.003609   0.049740 (  0.457492)


Using retry_errors: [Excon::Error::Timeout,Excon::Error::Socket,Excon::Error::Server]

  0.049035   0.003061   0.052096 (  0.422489)
```

I noticed this because we were using an old version of fog that predated the changes to retry configuration. We back ported just that configuration and observed significant performance regression on a couple of endpoints where the probability of a GET to an S3 object that didn't exist were high.

Excon sets the following for `retry_errors` https://github.com/excon/excon/blob/85556aeb4af10e94f876a8cbdb764f0377fa0d3a/lib/excon/constants.rb#L19-L23

```
  DEFAULT_RETRY_ERRORS = [
    Excon::Error::Timeout,
    Excon::Error::Socket,
    Excon::Error::HTTPStatus
  ]
```

Somewhat weirdly, the inheritance tree for successful responses has them as descendants of `Excon::Error`. https://github.com/excon/excon/blob/85556aeb4af10e94f876a8cbdb764f0377fa0d3a/lib/excon/error.rb#L68-L125

Incomplete tree for some additional context:
```
StandardError
└──Excon::Error
   ├──Warning
   ├──Socket
   │  └──Certificate
   ├──InvalidHeaderKey
   ├──InvalidHeaderValue
   ├──Timeout
   ├──ResponseParse
   ├──ProxyConnectionError
   ├──ProxyParse
   ├──TooManyRedirects
   └──HttpStatus
      ├──Informational
      ├──Success
      │  ├──OK
      │  ├──Created
      │  ├──Accepted
      │  └──[...]
      ├──Redirection
      ├──Client
      │  ├──BadRequest
      │  ├──Unauthorized
      │  ├──PaymentRequired
      │  ├──Forbidden
      │  ├──NotFound
      │  ├──MethodNotAllowed
      │  ├──NotAcceptable
      │  ├──ProxyAuthenticationRequired
      │  ├──RequestTimeout
      │  ├──Conflict
      │  ├──Gone
      │  └──[...]
      └──Server
         ├──InternalServerError
         ├──BadGateway
         ├──ServiceUnavailable
         └──GatewayTimeout
```   

It's the inclusion of `Excon::Error::Client` here that I question, almost all of this class of error seem like problems that won't be helped with a retry, perhaps with the exception of `RequestTimeout` and `TooManyRequests` and all of these will behave similarly to `NotFound`, causing performance to regress from <20ms to >6000ms.

I'll open up a PR here to patch the default configuration of `retry_errors` in this project, but it's certainly debatable whether the defaults being inherited are something that should be changed upstream in Excon itself.

I do also wonder on reflection whether `retry_limit: 5, retry_interval: 1` solved too narrowly for one use case - 1 second is an awfully long time in some contexts, particularly when response times for some S3 calls could be only a few milliseconds.

rahim:
Resolves #690

Excon's default configuration retries HTTP 4xx class errors, which we generally can't expect to be resolved by retry. Combined with relatively recent retry config added in https://github.com/fog/fog-aws/issues/674 we were seeing S3 calls to non-existent objects take multiple seconds when made via fog-aws.

See the issue for detailed explanation of the underlying cause.

Though this resolves the problem, I question whether it's the right approach:
- it's odd that the default retry behaviour (and connection options) for `Fog::AWS::Storage` are different to all the other services covered by fog-aws.
- ~~we create something of a foot gun, because someone who chooses to just configure eg `retry_limit` (or any other arbitrary connection option) will inadvertently set `retry_errors` back to Excon's default.~~ (addressed in later commit)
- possibly this is better addressed in Excon, I've opend https://github.com/excon/excon/issues/833 to ask that question there

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -554,7 +554,8 @@ module Fog
           @use_iam_profile = options[:use_iam_profile]
           @instrumentor       = options[:instrumentor]
           @instrumentor_name  = options[:instrumentor_name] || 'fog.aws.storage'
-          @connection_options = options[:connection_options] || DEFAULT_CONNECTION_OPTIONS
+          @connection_options =
+            DEFAULT_CONNECTION_OPTIONS.merge(options[:connection_options] || {})
           @persistent = options.fetch(:persistent, false)
           @acceleration = options.fetch(:acceleration, false)
           @signature_version = options.fetch(:aws_signature_version, 4)
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -14,6 +14,14 @@ module Fog
         'https' => 443
       }
 
+      DEFAULT_CONNECTION_OPTIONS = {
+        retry_limit: 5,
+        retry_interval: 1,
+        retry_errors: [
+          Excon::Error::Timeout, Excon::Error::Socket, Excon::Error::Server
+        ]
+      }
+
       MIN_MULTIPART_CHUNK_SIZE = 5242880
       MAX_SINGLE_PUT_SIZE = 5368709120
 
@@ -546,7 +554,7 @@ module Fog
           @use_iam_profile = options[:use_iam_profile]
           @instrumentor       = options[:instrumentor]
           @instrumentor_name  = options[:instrumentor_name] || 'fog.aws.storage'
-          @connection_options     = options[:connection_options] || { retry_limit: 5, retry_interval: 1 }
+          @connection_options = options[:connection_options] || DEFAULT_CONNECTION_OPTIONS
           @persistent = options.fetch(:persistent, false)
           @acceleration = options.fetch(:acceleration, false)
           @signature_version = options.fetch(:aws_signature_version, 4)


rahim:
> we create something of a foot gun, because someone who chooses to just configure eg retry_limit (or any other arbitrary connection option) will inadvertently set retry_errors back to Excon's default.

Across fog-aws we have the pattern `@connection_options = options[:connection_options] || {}`

```
lib/fog/aws/efs.rb:75:          @connection_options = options[:connection_options] || {}
lib/fog/aws/elb.rb:133:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/beanstalk.rb:68:          @connection_options = options[:connection_options] || {}
lib/fog/aws/cdn.rb:148:          @connection_options = options[:connection_options] || {}
lib/fog/aws/cloud_watch.rb:90:          @connection_options = options[:connection_options] || {}
lib/fog/aws/sts.rb:81:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/redshift.rb:81:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/cloud_formation.rb:66:          @connection_options = options[:connection_options] || {}
lib/fog/aws/support.rb:90:          @connection_options = options[:connection_options] || {}
lib/fog/aws/data_pipeline.rb:107:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/dns.rb:100:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/ses.rb:55:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/emr.rb:63:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/glacier.rb:187:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/lambda.rb:85:          @connection_options = options[:connection_options] || {}
lib/fog/aws/auto_scaling.rb:96:          @connection_options = options[:connection_options] || {}
lib/fog/aws/kinesis.rb:39:          @connection_options = options[:connection_options] || {}
lib/fog/aws/kms.rb:94:          @connection_options = options[:connection_options] || {}
lib/fog/aws/storage.rb:549:          @connection_options     = options[:connection_options] || { retry_limit: 5, retry_interval: 1 }
lib/fog/aws/iam.rb:264:          @connection_options = options[:connection_options] || {}
lib/fog/aws/sns.rb:83:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/sqs.rb:81:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/ecs.rb:62:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/dynamodb.rb:80:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/rds.rb:217:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/federation.rb:39:          @connection_options = options[:connection_options] || {}
lib/fog/aws/simpledb.rb:74:          @connection_options     = options[:connection_options] || {}
lib/fog/aws/compute.rb:554:          @connection_options           = options[:connection_options] || {}
```

That generally makes, if connection options aren't set, ensure we pass an empty hash to Excon.

For the custom retry settings that have been applied only to `Storage`, we'd be better to merge any specified connection options over the defaults. I'll make that change.

geemus:
I hope I covered it relatively clearly in the other issues, but just in case I'll briefly say that the idea (if I recall) was to make it easier to deal with the eventually consistent nature of AWS. ie if you create an object and then want to address it, you'd like that to just work even if it takes a little bit longer. That said, cases like what you point out show how that was inadvertently applied much too broadly and by seeking to make that case easier it introduces a lot of lag for other use cases. I'm not yet sure what the best approach or solution really is, but tried to leave some ideas and ask some questions across the issues, so I hope we'll be able to talk through them and figure something out. That being said, I'd like to chat a bit more before proceeding with this or similar patches to make sure we've considered the options and impacts thoroughly. Thanks.

geemus (approved):
After our discussions, I believe this should fix issues that some folks were having and not negatively impact the eventual consistency things it was originally set out to address (as AWS seems to have fixed that on their side of things).

geemus:
@rahim did you need a release with this?

rajyan:
Thank you very much for looking into this issue and thank you for your swift responses! 

I’m not the author of this PR, but I think this issue reported in carrierwave seems related
https://github.com/carrierwaveuploader/carrierwave/issues/2697
so I would be grateful if you could release it.

geemus:
@rajyan Thanks, I hadn't seen that issue, though I'm not certain it will totally fix it either (it should hopefully at least help).

Released in v3.21.0.

PR #703, opened 2024-03-12
Add ca-west-1 region

### LINKED ISSUE #701: Missing AWS Region - ca-west-1 (Calgary)
Hello,

We noticed that Fog does not have the relatively new ca-west-1 (Calgary) region. Does it just need to be added to the [aws.rb](https://github.com/fog/fog-aws/blob/41e341f5a3d19ea9ff60e67587554db6253ba6e7/lib/fog/aws.rb#L221-L241) file? Or are the additional changes needed in the library for new regions?

jgrammen-agilitypr:
Add ca-west-1 region to region list in aws.rb
This is an attempt to fix the missing region (based on comments in #701 ), my live testing suggests it works, but I am inexperienced with ruby programming.

 fixes #701

### DIFF
--- a/lib/fog/aws.rb
+++ b/lib/fog/aws.rb
@@ -225,7 +225,7 @@ module Fog
         'ap-northeast-1', 'ap-northeast-2', 'ap-northeast-3',
         'ap-south-1',
         'ap-southeast-1', 'ap-southeast-2', 'ap-southeast-3', 'ap-southeast-4',
-        'ca-central-1',
+        'ca-central-1', 'ca-west-1',
         'cn-north-1',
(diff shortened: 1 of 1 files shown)


geemus (approved):
LGTM, thanks!

geemus:
Note: I'm not sure what's up with the actions/tests. The "expected" stuff is all the old versions, which I thought I recently disabled in testing to focus on non-EOL rubies. In any event, I don't think we need to block on this here.

geemus:
Released in v3.21.1.

jgrammen-agilitypr:
@geemus thank you for the new version. 👍

PR #707, opened 2024-04-02
Fix typos

kianmeng:
Found via `typos --format brief`

### DIFF
--- a/lib/fog/aws/requests/elasticache/create_cache_cluster.rb
+++ b/lib/fog/aws/requests/elasticache/create_cache_cluster.rb
@@ -19,7 +19,7 @@ module Fog
         #   * :engine_version <~String> - The Cluster's caching software version
         #   * :notification_topic_arn <~String> - Amazon SNS Resource Name
         #   * :port <~Integer> - The memcached port number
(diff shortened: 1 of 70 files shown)


lib/fog/aws/requests/elasticache/create_cache_cluster.rb
```diff
@@ -19,7 +19,7 @@ class Real
         #   * :engine_version <~String> - The Cluster's caching software version
         #   * :notification_topic_arn <~String> - Amazon SNS Resource Name
         #   * :port <~Integer> - The memcached port number
-        #   * :preferred_availablility_zone <~String>
+        #   * :preferred_availabilility_zone <~String>
```
geemus, line 22: This correction looks off, I think it needs one fewer `il` in the middle? That should match with the other change below and the related one in the other file above.

geemus (changes requested):
Thanks for all the fixes. I think one of the corrections was inaccurate, but otherwise looking good.

geemus (approved):
Thanks!

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #711, opened 2024-06-15
Fix frozen string literal issues part 2

chaadow:
Followup: https://github.com/fog/fog-aws/pull/709

After running tests locally I managed to fix all of the frozen string literals issues. 

This PR however depends on `fog-core` and `fog-xml` being fixed ( I will create separate PRs on those repos)

It's worth mentioning that on fog-core and fog-xml, I will just fix just enough to make the tests green her
I haven't yet run the tests on those repos.

@geemus What do you think? Do you mind providing me with a simple github action .yml file in which i can enable RUBYOPT

here is the command I run locally: `bundle exec rake RUBYOPT="--enable-frozen-string-literal"`


chaadow:
Ok sorry, I managed to find the shared config ( i was a bit lost with `.github/.github` thought it was some Github magic)

As we've said in the previous PR, the idea is to manually enable it gem by gem, and once we feel comfortable we make it common to every gem in the shared CI config.

(Right now, tests are failing because the dependent PRs are not yet merged/released)

chaadow:
I think a good idea would be to add timeouts on the CI, it's taking a bit long i'm sorry. I will do tomorrow

PR #715, opened 2024-06-27
Prefer AWS_REGION over AWS_DEFAULT_REGION in credential fetching

stanhu:
As https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html#feature-region-sdk-compat describes, the AWS CLI v2 uses AWS_REGION before AWS_DEFAULT_VERSION.

### DIFF
--- a/lib/fog/aws/credential_fetcher.rb
+++ b/lib/fog/aws/credential_fetcher.rb
@@ -21,7 +21,7 @@ module Fog
           if options[:use_iam_profile]
             begin
               role_data = nil
-              region = options[:region] || ENV["AWS_DEFAULT_REGION"]
+              region = options[:region] || ENV["AWS_REGION"] || ENV["AWS_DEFAULT_REGION"]
 
               if ENV["AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"]
                 connection = options[:connection] || Excon.new(CONTAINER_CREDENTIALS_HOST)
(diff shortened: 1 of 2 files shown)


geemus (approved):
LGTM, thanks!

geemus:
@stanhu did you need a release?

stanhu:
@geemus Sure, thanks!

lib/fog/aws/credential_fetcher.rb
```diff
@@ -21,7 +21,7 @@ def fetch_credentials(options)
           if options[:use_iam_profile]
             begin
               role_data = nil
-              region = options[:region] || ENV["AWS_DEFAULT_REGION"]
+              region = options[:region] || ENV["AWS_REGION"] || ENV["AWS_DEFAULT_REGION"]
```
stanhu, line 24: Actually, I question whether the environment is ever used because `options[:region]` appears to default to `us-east-1` in https://github.com/fog/fog-aws/blob/b96b27a6bc5d8ada0672abd98a2de330c99b3426/lib/fog/aws/storage.rb#L298-L304?

stanhu: Never mind, I confirmed that this works properly if you use `Fog::Storage.new` since fog-core appears to call `fetch_credentials` directly from `handle_settings`.

geemus:
Apologies for delay, now released in v3.24.0.

PR #717, opened 2024-08-07
Drop warning if region is not us-east-1 in mock

stanhu:
Previously if `Fog.mock!` were used with some other region and the bucket already existed, a warning message, `Your region <region> does not match the default region 'us-east-1'`.

We were seeing a lot of these warnings in our test suite. As far as I can tell, there's no reason why some other region can't be used, so drop this.

### DIFF
--- a/lib/fog/aws/requests/storage/put_bucket.rb
+++ b/lib/fog/aws/requests/storage/put_bucket.rb
@@ -61,10 +61,6 @@ DATA
           end
           if !self.data[:buckets][bucket_name]
             self.data[:buckets][bucket_name] = bucket
-          elsif self.region != 'us-east-1'
-            response.status = 409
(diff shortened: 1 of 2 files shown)


stanhu:
@geemus I think this logic has been in place for a long time, but I'm not sure why. Is this just to ensure that tests only use one well-defined region? I suppose tests could fail if buckets with the same name but different regions are used.

geemus:
@stanhu good question. Presumably we had a reason at the time, but I certainly don't remember it now. I agree it seems pretty arbitrary and although it could lead to some issues with re-used bucket names it's not clear that is any worse than the current confusing and probably not very true-to-real mock behavior.

geemus (approved):
There was probably a reason for this, but it's lost to time. So let's go with the simpler/less confusing version.

geemus:
Released in v3.25.0.

PR #719, opened 2024-08-10
Disable S3 Signature v4 Streaming by default

stanhu:
The STREAMING-AWS4-HMAC-SHA256-PAYLOAD streaming method is not widely implemented in AWS SDKs, and many third-party S3 implementations don't support this.

S3 multipart uploads are a more standard way of handling chunked transfers in any case.


geemus:
@stanhu Does this need a release?

geemus:
Released in v3.25.0.

PR #727, opened 2024-10-14
Update SES to use AWS SigV4

### LINKED ISSUE #726: SES needs to support AWS SigV4
https://github.com/fog/fog-aws/blob/b22147369ab2f376efcb5c713233c9e53e288e46/lib/fog/aws/ses.rb#L90-L95

Sadly it seems like SES is the only service left behind (ok, maybe also DNS) that doesn't support SigV4, and so we get this message from AWS:

```
Signature Version 3 requests are deprecated from March 1, 2021. From that date on, we are progressively rejecting such requests. To resolve the issue you must migrate to Signature Version 4. If you are self-signing your requests, refer to the documentation for Authenticating requests to the Amazon SES API [1] with Signature Version 4 [2]. If you are not self-signing your requests, simply update your SDK/CLI to the latest version. [1] https://docs.aws.amazon.com/ses/latest/DeveloperGuide/using-ses-api-authentication.html [2] https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
```

ohookins:
Resolves #726 

Have tested it works.

### DIFF
--- a/lib/fog/aws/ses.rb
+++ b/lib/fog/aws/ses.rb
@@ -48,18 +48,21 @@ module Fog
         def initialize(options={})
 
           @use_iam_profile = options[:use_iam_profile]
-          setup_credentials(options)
 
           @instrumentor       = options[:instrumentor]
           @instrumentor_name  = options[:instrumentor_name] || 'fog.aws.ses'
           @connection_options     = options[:connection_options] || {}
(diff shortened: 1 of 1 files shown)


geemus (approved):
LGTM, thanks!

geemus:
released in v3.28.0

PR #730, opened 2024-11-18
Support FIPS endpoints, disable S3 Transfer Acceleration in GovCloud

matt-domsch-sp:
The Fog::AWS::Utils region_to_host method returns the standard S3 endpoints even when ENV['AWS_USE_FIPS_ENDPOINT']=='true'.  When FIPS is called for, and we are in a region where FIPS endpoints are available,  this method should return the FIPS endpoint.

Furthermore, when S3 Transfer Acceleration (S3TA) is requested by configuration, the above endpoint gets overridden to select the S3TA endpoint.  However, S3TA is not avaialble in GovCloud, and has no FIPS endpoint equivalents.  In this instance, if the region is a GovCloud region, or if FIPS mode is called for, do _not_ override the endpoint to use S3TA.

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -7,6 +7,8 @@ module Fog
 
       DEFAULT_REGION = 'us-east-1'
       ACCELERATION_HOST = 's3-accelerate.amazonaws.com'
+      AWS_FIPS_REGIONS = ['us-east-1', 'us-east-2', 'us-west-1', 'us-west-2', 'us-gov-east-1', 'us-gov-west-1', 'ca-central-1', 'ca-west-1'].freeze
+      AWS_GOVCLOUD_REGIONS = ['us-gov-east-1', 'us-gov-west-1'].freeze
 
       DEFAULT_SCHEME = 'https'
       DEFAULT_SCHEME_PORT = {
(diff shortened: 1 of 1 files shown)


geemus:
@matt-domsch-sp Thanks for all the details and the patch. The code looks reasonable to me, though I have zero experience with govcloud, fips, and acceleration. Is there anything else we should check or test before proceeding, or did you have questions? (I noticed it's marked as work-in-progress still)

matt-domsch-sp:
I'm not aware of anything else.  There aren't any tests for this code and I haven't added any either.  It would be a behavior change re no S3TA for systems that enable the FIPS mode flag, but that's probably few and far between.  I've marked it ready for review.

geemus:
@matt-domsch-sp Thanks. We could consider having an error or at least warning in the case where you are trying to enable acceleration in a context where it would be unavailable. That might help smooth this out, though I presume it wouldn't have worked previously anyway. What do you think?

matt-domsch-sp:
I've added a warning and explicit disabling of acceleration. @geemus thanks for the suggestion.

lib/fog/aws/storage.rb
```diff
@@ -578,6 +584,13 @@ def initialize(options={})
             @port       = options[:port]        || DEFAULT_SCHEME_PORT[@scheme]
           end
 
+          # GovCloud doesn't support S3 Transfer Acceleration https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-s3.html
+          # S3 Transfer Acceleration doesn't support FIPS endpoints.  When both fog_aws_accelerate=true and AWS_USE_FIPS_ENDPOINT=true, don't use Accelerate.
+          if @acceleration && (AWS_GOVCLOUD_REGIONS.include?(@region) || ENV['AWS_USE_FIPS_ENDPOINT'] == 'true')
+            Fog::Logger.warning("fog: S3 Transfer Acceleration is not available in GovCloud regions or when AWS_USE_FIPS_ENDPOINT=true. Disabling accelearation.")
```
geemus, line 590:
```suggestion
            Fog::Logger.warning("fog: S3 Transfer Acceleration is not available in GovCloud regions or when AWS_USE_FIPS_ENDPOINT=true. Disabling acceleration.")
```

geemus (approved):
Looks like a typo snuck into the warning message, but otherwise looking good. Thanks!

Issue #738, opened 2025-05-07
Extremely slow transfer rate to S3


avoidik:
Hello,

Could you assist me with the following concern? At the moment I am observing an extremely slow transfer rate to S3 when using fog-aws library. I have a 1,4TB archive of compressed random data (with low entropy). Whenever I transfer that file using awscli the process is completed in 22 minutes, but when I do this using the fog-aws library, it takes 6 hours.

I would be happy to get any suggestion in order to troubleshoot, identify, and address the root cause.

Regards

geemus:
@avoidik Thanks for the detailed report. I don't know for sure, but can make some educated guesses (especially given the other message you shared about chunk count and size). I think generally, fewer/larger chunks will be faster. I think it defaults to the largest supported size, so leaving it at that would likely help (if you aren't already doing that).

Could you share the code you are using to do the upload with fog-aws and the command line options you are using with awscli? I think that may help me better understand discrepancies and hopefully start to narrow in on what might be happening.

avoidik:
> fewer/larger chunks will be faster

@geemus I might be wrong in my analysis, please correct me if so

in the current implementation every `multipart_chunk_size` chunk will be cached in memory. With `multipart_chunk_size = 100MB` and `concurrency = 10` the process will consume at least 1000MB of RAM. If I set either of these two properties too large I'll increase the odds of OOM failure. On contrary, if I'm not going to set the `multipart_chunk_size` at all, wouldn't it be changed to `MAX_SINGLE_PUT_SIZE = 5368709120` (5GB) for files larger than 5GB? Meaning that 5GB RAM will be required.

Based on:
https://github.com/fog/fog-aws/blob/v3.31.0/lib/fog/aws/storage.rb#L25
https://github.com/fog/fog-aws/blob/v3.31.0/lib/fog/aws/storage.rb#L210-L213
https://github.com/fog/fog-aws/blob/v3.31.0/lib/fog/aws/models/storage/file.rb#L278-L282

> Could you share the code you are using to do the upload with fog-aws and the command line options you are using with awscli?

I have been using hyperfine to benchmark transfer performance. Here's a summary:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `measure-transfer-rate-fog-5mb-1` | 379.257 ± 1.955 | 377.288 | 381.198 | 33.05 ± 0.92 |
[...]

PR #739, opened 2025-05-13
add concurrency to multipart_save method

avoidik:
fix for https://github.com/fog/fog-aws/issues/738

### DIFF
--- a/lib/fog/aws/models/storage/file.rb
+++ b/lib/fog/aws/models/storage/file.rb
@@ -328,17 +328,33 @@ module Fog
           # Store ETags of upload parts
           part_tags = []
 
+          # Calculate total size and ensure we don't exceed part limit
+          total_size = Fog::Storage.get_body_size(body)
+          parts_count = (total_size.to_f / multipart_chunk_size).ceil
+
+          # AWS S3 has a hard limit of 10,000 parts, make sure we are below this limit for large objects
+          if parts_count > 10000
+            self.multipart_chunk_size = (total_size.to_f / 10000).ceil
+            parts_count = 10000
+          end
+
           # Upload each part
-          # TODO: optionally upload chunks in parallel using threads
-          # (may cause network performance problems with many small chunks)
-          # TODO: Support large chunk sizes without reading the chunk into memory
           if body.respond_to?(:rewind)
-            body.rewind  rescue nil
+            body.rewind rescue nil
           end
-          while (chunk = body.read(multipart_chunk_size)) do
-            part_upload = service.upload_part(directory.key, key, upload_id, part_tags.size + 1, chunk, part_headers(chunk))
-            part_tags << part_upload.headers["ETag"]
-          end
+
+          pending = PartList.new(
+            (1..parts_count).map do |part_number|
+              UploadPartData.new(part_number, {}, nil)
+            end
+          )
+          thread_count = self.concurrency
+          completed = PartList.new
+          errors = upload_parts_in_threads(directory.key, key, upload_id, pending, completed, thread_count)
+
+          raise errors.first if errors.any?
+
+          part_tags = completed.to_a.sort_by { |part| part.part_number }.map(&:etag)
 
           if part_tags.empty? #it is an error to have a multipart upload with no parts
             part_upload = service.upload_part(directory.key, key, upload_id, 1, '', part_headers(''))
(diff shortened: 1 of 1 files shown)


lib/fog/aws/models/storage/file.rb
```diff
@@ -328,18 +328,34 @@ def multipart_save(options)
...
+          completed = PartList.new
+          errors = upload_parts_in_threads(directory.key, key, upload_id, pending, completed, thread_count)
+
+          raise errors.first if errors.any?
+
+          part_tags = completed.to_a.sort_by { |part| part.part_number }.map(&:etag)
+
           if part_tags.empty? #it is an error to have a multipart upload with no parts
             part_upload = service.upload_part(directory.key, key, upload_id, 1, '', part_headers(''))
             part_tags << part_upload.headers["ETag"]
```
avoidik, line 361: there is no similar logic in `multipart_copy()` method 🤔

lib/fog/aws/models/storage/file.rb
```diff
@@ -460,6 +476,49 @@ def upload_in_threads(target_directory_key, target_file_key, upload_id, pending,
...
+              begin
+                while part = pending.shift
+                  # Determine byte range for this part
+                  start_pos = (part.part_number - 1) * multipart_chunk_size
+
+                  # Safely read the chunk from body
+                  chunk = nil
+                  mutex.synchronize do
+                    if body.respond_to?(:seek)
+                      body.seek(start_pos) rescue nil
```
avoidik, line 495: hello, it is me, a little bottleneck right here!

lib/fog/aws/models/storage/file.rb
geemus: Probably an oversight I guess?

lib/fog/aws/models/storage/file.rb
geemus: Yeah. This should prevent multiple threads from trying to read the file at the same time I guess. They would take turns leading it into memory, but then actually do the upload in parallel (which is the slower part). I think mutexing the file read is probably necessary if I recall properly. Is that what you were thinking? Happy to discuss further.

avoidik: Yes, this will definitely be a bottleneck on a block storage side. Multiple threads trying to seek through blocks and run IO operations from one file, that is a reason why mutex is there.

geemus: Any ideas for alternatives or workarounds?

avoidik: in this implementation it's implied that we could fine-tune this bottleneck using two existing properties, `concurrency` and `multipart_chunk_size`

avoidik: Using some sort of buffering in addition to threads wouldn't make it any better.

geemus:
@avoidik - This seems like a pretty reasonable next step to try to me. Have you had a chance to test it? Were there additional parts you wanted to change before we proceed?

avoidik:
@geemus I didn't test it thoroughly, I could run yet another performance benchmark again after your minimal input and some kind of a review

geemus:
@avoidik Yeah, that sounds good. Overall this code appears to align well with the copy code that was already here, but I don't have direct experience using that either. It will also be helpful to know if and how much it helps with the performance issues you originally noted. Thanks!

lib/fog/aws/models/storage/file.rb
```diff
@@ -328,18 +328,34 @@ def multipart_save(options)
           # Store ETags of upload parts
           part_tags = []
 
+          # Calculate total size and ensure we don't exceed part limit
+          total_size = Fog::Storage.get_body_size(body)
+          parts_count = (total_size.to_f / multipart_chunk_size).ceil
+
+          # AWS S3 has a hard limit of 10,000 parts, make sure we are below this limit for large objects
+          if parts_count > 10000
```
aslafy-z, line 336: This is a AWS limit but some other providers have lower limits, eg: Scaleway has a 1000 parts limit https://www.scaleway.com/en/docs/object-storage/faq/#how-can-i-upload-large-objects

It would be awesome if the maximum value was a configurable value, like they did in rclone: https://github.com/rclone/rclone/commit/7458d37d2a083d53f7d3847fcd8e581661ce63b7

geemus: Good callout. I'd welcome a PR that makes this adjustable if you'd like to take a shot at it. Thanks!

PR #742, opened 2025-05-16
Add ability to specify s3 restore object tier in post_object_restore

### LINKED ISSUE #740: Add ability to specify s3 restore object tier in post_object_restore
- https://github.com/fog/fog-aws/blob/master/lib/fog/aws/requests/storage/post_object_restore.rb
- https://docs.aws.amazon.com/AmazonS3/latest/API/API_RestoreObject.html#API_RestoreObject_RequestSyntax
- Body can be something like this:
```
            <RestoreRequest xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
              <Days>#{days}</Days>
              <GlacierJobParameters>
                <Tier>#{tier}</Tier>
              </GlacierJobParameters>
            </RestoreRequest>
```
- We can just add another optional parameter for `tier`

I can make a PR if this is ok.

Ankk98:
Fix #740

### DIFF
--- a/lib/fog/aws/requests/storage/post_object_restore.rb
+++ b/lib/fog/aws/requests/storage/post_object_restore.rb
@@ -7,6 +7,7 @@ module Fog
         # @param bucket_name [String] Name of bucket containing object
         # @param object_name [String] Name of object to restore
         # @option days [Integer] Number of days to restore object for. Defaults to 100000 (a very long time)
+        # @option tier [String] Glacier retrieval tier. Can be 'Standard', 'Bulk', or 'Expedited'
         #
         # @return [Excon::Response] response:
         #   * status [Integer] 200 (OK) Object is previously restored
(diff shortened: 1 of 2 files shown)


lib/fog/aws/requests/storage/post_object_restore.rb
```diff
@@ -15,11 +16,16 @@ class Real
         #
         # @see http://docs.amazonwebservices.com/AmazonS3/latest/API/RESTObjectPOSTrestore.html
         #
-        def post_object_restore(bucket_name, object_name, days = 100000)
+        def post_object_restore(bucket_name, object_name, days = 100000, tier = nil)
```
geemus, line 19: Rather than adding another argument here, I think it might be a good idea to covert this to using an options hash (which could then include both days and tier, but also allow for future compatibility with additional options).

You can see an example of how we've done that conversion on other methods here: https://github.com/fog/fog-aws/blob/c84f6aa666a2626bca308082510f31d3e7e6b0b8/lib/fog/aws/requests/compute/create_volume.rb#L32-L36

Does that make sense? Otherwise everything here looks good, but I think going directly to that style will save us some trouble in the long run.

geemus (changes requested):

geemus (approved):
Looks great, thanks for updating that to match other options style methods.

geemus:
I'm traveling for the next few days, so I'd rather not release just now, but will plan to do so early next week. Do feel free to ping me if you need it sooner or I haven't yet gotten to it by later in the week. Thanks!

PR #755, opened 2026-04-10
bug(aws/storage): include non-standard port in Host header for signature v4

stanhu:
Per the AWS SigV4 spec and RFC 7230, the Host header must include the port when it is not the default for the scheme. The regular request path was always setting Host to just the hostname, causing signature verification failures against S3-compatible servers (e.g. HCP) on non-standard ports. The presigned URL path already handled this correctly; align the request path to match.

### DIFF
--- a/lib/fog/aws/storage.rb
+++ b/lib/fog/aws/storage.rb
@@ -658,8 +658,9 @@ module Fog
           params = request_params(params)
           scheme = params.delete(:scheme)
           host   = params.delete(:host)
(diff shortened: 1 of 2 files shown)


geemus (approved):
LGTM, thanks.

stanhu:
@geemus Thanks! Could you release a version for me?

geemus:
@stanhu thanks, released in v3.33.2.

PR #759, opened 2026-08-27
Fix thread safety of shared Fog::Storage connections

stanhu:
## Problem

`Fog::AWS::Storage::Real` is not safe to share across threads. This surfaced in a GitLab investigation of a process-wide cached connection behind Puma. There are two races in the request path, both triggered when a single connection is used concurrently.

### Credential refresh race

`#request` read `@aws_session_token` early and `@signer` about 35 lines later, as two separate reads. Under `use_iam_profile`, `refresh_credentials` → `setup_credentials` replaces four correlated ivars together: the access key id, secret, session token, and a rebuilt signer. A refresh landing between the two reads produces a request carrying generation N's session token signed by generation N+1's signer, which S3 rejects with a 403 (`SignatureDoesNotMatch` / `InvalidToken`).

Wrapping the writes in a mutex does not fix this; the reader's window is what has to close.

Affects installs using IAM instance profiles / ECS task roles / EKS IRSA (`use_iam_profile: true`). Static-key installs never refresh and are unaffected.

### Redirect race

`#_request` rescued an S3 301/307 redirect and, to retry against the correct region, swapped the shared `@region` and `@signer`, issued a nested request, then restored them. Any concurrent request on the same connection could observe the swapped-in region mid-flight. This one is independent of `use_iam_profile`.

## Fix

Bundle the correlated credential state into a single immutable `Credentials` object held in `@credentials`. `setup_credentials` builds a new frozen one and swaps the reference in a single assignment; readers grab it once into a local, so a concurrent refresh can never tear the `(token, signer)` pair. This is lock-free and mirrors how `Google::Auth` avoids the same problem (single ivar read into a local).

- `#request` snapshots `@credentials` once, then hands it to a new private `perform_request`, which does all signing.
- On a redirect, `#_request` takes a fresh snapshot, builds a signer for the new region locally, and re-dispatches through `perform_request` instead of mutating `@region`/`@signer`.
- `#request` and `#_request` keep their existing signatures.
- The presigned-URL helpers, `signature_v2`, `#cdn`, the `Mock` data helpers, and `post_object_hidden_fields` now read the same snapshot.

## Testing

Added `tests/storage_thread_safety_tests.rb`, which forces the interleave (a `Thread.pass` inside the existing reader window) and asserts zero torn `(token, signer)` pairs across 4,000 concurrent requests, plus a guard that refreshes actually interleaved. A negative control reproducing the old separate-read pattern produced torn reads, confirming the test catches the bug.


geemus (approved):
lgtm, thanks!

geemus:
@stanhu do you need me to cut a release as well?

stanhu:
@geemus Yes, thank you SO much!

geemus:
@stanhu k. I'll plan to do one tomorrow morning as I'm about to wrap for the day.

geemus:
released in v3.33.3

stanhu:
Thanks again!

PR #760, opened 2026-08-27
Fix flaky describe_spot_price_history mock

stanhu:
## Problem

`tests/requests/compute/spot_price_history_tests.rb` fails intermittently (~1 in 1000 runs) with:

```
InvalidParameterValue => Invalid value '0' for maxResults (Fog::AWS::Compute::Error)
  lib/fog/aws/requests/compute/describe_spot_price_history.rb:92:in 'Fog::AWS::Compute::Mock#describe_spot_price_history'
```

When the caller passes no `MaxResults`, the mock generates a default with `Fog::Mock.random_numbers(3).to_i`. `random_numbers(3)` can return `"000"`, which `.to_i` turns into `0`, tripping the mock's own `max_results > 0` validation and raising.

## Fix

Clamp the generated default to a minimum of `1`. Validation of a caller-supplied `MaxResults` is unchanged.

## Testing

- `tests/requests/compute/spot_price_history_tests.rb` passes.
- Forcing `Fog::Mock.random_numbers` to return `"000"` raises the exact CI error before this change and returns normally after it.


stanhu:
@geemus Would you mind reviewing this and #759?

geemus (approved):
lgtm, thanks!

PR #761, opened 2026-09-19
Let the connection set the concurrency of multipart copies

mshibuya:
## Problem

A multipart copy copies its parts one at a time unless the file is told to use more threads:

```ruby
file.concurrency = 10
file.copy(target_bucket, target_key)
```

That only works if you are the one holding the file. When a library copies on your behalf, you have nowhere to put it. CarrierWave builds the file inside its own storage layer, so GitLab patches CarrierWave just to set this: [carrierwave_fog_acl_patch.rb#L76](https://github.com/gitlabhq/gitlabhq/blob/b70c33f3d27d5679e8c8edf068136b5a18f5c791/config/initializers/carrierwave_fog_acl_patch.rb#L76).

The chunk size that decides whether a copy is multipart at all doesn't have this problem: `max_copy_chunk_size` is a connection option.

## Change

`copy_concurrency` is now a connection option too, and `File#concurrency` falls back to it:

```ruby
storage = Fog::Storage.new(
  provider: 'AWS',
  max_copy_chunk_size: 10 * 1024 * 1024,
  copy_concurrency: 10,
  ...
)
```

A file given its own `concurrency` still uses that one. The default is unchanged at 1, so nothing copies differently unless the option is set. Values below 1 raise `ArgumentError`, as `File#concurrency=` already does.

## Tests

Four tests in `tests/requests/storage/multipart_copy_tests.rb`: the default, the value from the connection, a file overriding it, and the error. The storage tests pass (285 succeeded, up from 281; 12 pending either way).


geemus:
@mshibuya This looks/sounds good, thanks for pointing it out.

I have one request and one question. First, could you rebase this on the latest main? (there is an unrelated CI-break that just surfaced and I fixed after you created this). Second, I wonder if we should be more specific in the key name. I think what you have might actually be just fine, but I could also see it being a longer, more specific name for clarity. I could go either way, I think it's probably fine as-is, but thought I would mention it while I'm thinking about it.

mshibuya:
Thanks for the review! Rebased to the master.

About the name, I'd lean towards keeping the current name `copy_concurrency`, as it pairs with the existing `max_copy_chunk_size`: both only take effect on multipart copies, and they tune the same operation. A longer name like `multipart_copy_concurrency` would be more precise, but then the two would read differently.

Happy to change it if there's any recommendation 👍

geemus:
@mshibuya - thanks for that. Yeah, my inclination on the naming was the same as yours, but I wanted to at least mention it so I could get your opinion too. I'll take another look now, but I think we should be good as long as CI lands. Will you need a release?

geemus (approved):
lgtm, thanks!

geemus:
@mshibuya also, would be great to unwind the gitlab patch if they don't need it any more.

mshibuya:
Thank you!
Of course, that's a part of my effort to upgrade their CarrierWave dependency 👍

